PR #59, opened 2016-01-27
[WIP] Hyperparameter search scripts

evanfeinberg:
For NNScore features.


evanfeinberg:
Gently tweaked the API to accommodate (1) model creation by passing instances of classes rather than ```model_type``` string and (2) allow the user to use virtually any ```sklearn``` model in addition to ```deepchem``` models.

PR #128, opened 2016-03-23
CoulombTransformers and Dataset

joegomes:
This PR adds:
- CoulombRandomization transformer for randomizing coulomb matrices
- CoulombBinarization transformer for binarizing coulomb matrices
- update_moments() added to DataSet object to re-calculate statistics on transformed data
- Fixed all transformer classes to read from transformed data by default


joegomes:
Tests pass, merging.

PR #141, opened 2016-03-29
Safe normalization for CoulombBinarizationTransformer

joegomes:
This PR adds:
- A safe normalization routine for the `CoulombBinarizationTransformer` that is automatically performed after binarization.

The existing `NormalizationTransformer` (and related stats code) contains a divide by zero bug if there exists a column within X where X[:,i] = mean(X[:,i]) and std(X[:,i]) = 0.  We first found this bug when stacking normalization following `CoulombBinarizationTransformer`.  Switching to the np mean and std functions give the correct normalization.

@rbharath Instead of merging right away, we could use this PR to discuss whether to change the statistics functions in `Dataset` to something similar to what I've done in `transform`.  It could make sense to combine `get_statistics` and `update_moments` into one function that loads X-transformed from disk and computes the stats, as well as get rid of X_sums and X_squares from the metadata df (I think this is the only time when these are calculated or used).

### DIFF
--- a/deepchem/transformers/__init__.py
+++ b/deepchem/transformers/__init__.py
@@ -50,7 +50,6 @@ class Transformer(object):
 
     Adds X-transform, y-transform columns to metadata.
     """
-    dataset.update_moments()
     df = dataset.metadata_df
     indices = range(0, df.shape[0])
     transform_row_partial = partial(_transform_row, df=df, transformer=self)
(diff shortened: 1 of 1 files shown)


joegomes:
After discussing IRL, @rbharath and I decided to keep the changes to `CoulombBinarizationTransformer` and leave `get_statistics` alone for now.  Merging #141

PR #337, opened 2016-12-28
[WIP] Add GDB7 model to benchmarks

joegomes:
This PR adds:
- A kernel ridge regression model on GDB7 atomization energies dataset for Deepchem benchmarks

Current work in progress:
- Fix KRR performance (worse compared with similar models created Spring 2016)
- Fixes for `CoulombMatrixRandomization` and `CoulombMatrixBinarization` transformers
- A tensorflow model on GDB7 atomization energies dataset


rbharath:
@joegomes Thanks for sending this in! Would you be able to remove the data files from this PR and put them all on S3?  (I think I'm going to start replacing all datasets in deepchem with shell scripts so the github source stays small.)

peastman:
>  (I think I'm going to start replacing all datasets in deepchem with shell scripts so the github source stays small.)

Unfortunately it's too late for that.  Once a file is added to the repository, it's there forever, and because of the broken way git is designed, every person who checks out the code gets a complete copy of the repository on their computer, and hence a copy of every file that has ever been committed.  At this point, the only way I know of to fix it is to create a new repository from scratch with a clean copy of only the files you want to have.

Even in the case of this PR, if you just add a new commit removing the files, they'll still be present in the history.  You either need to do a rebase to eliminate any revision that ever contained them, or else do a new PR from a clean branch.

rbharath:
@peastman Is that necessarily true? There seem to be repo-cleaner tools for
doing hard-purges https://rtyley.github.io/bfg-repo-cleaner/

I figure we can just use one of these tools to hard remove the datasets in
the history.

PR #396, opened 2017-02-14
pdbbind datasets blacklist

lilleswing:
[full_grid.tar.gz](https://github.com/deepchem/deepchem/files/774427/full_grid.tar.gz)

Some pdb's cause OpenBabel to segfault attempting to load them.  I did a fork/exec loop to figure out which PDB's they were.  Here is the final black list and the featurized joblib files.  

```
TIMING: PDBBind Featurization took 16560.041 s // 4.60001138889 hours
```


proteneer:
nice - did you guys ever figure out what was causing the segfaults to happen?

lilleswing:
Openbabel was attempting to kekulize the molecule while loading into an OpenBabel mol object.  I was able to repro it with a debugger attached, but decided to just ignore the proteins instead of digging into the OpenBabel kekulize source.

rbharath:
@lilleswing Thanks for the PR! Looking through the code, it looks like the pi-pi stacking/cation-pi bonds haven't been turned on for `GridFeaturizer`. (lines 83-84 in `pdbbind_datasets.py`). Would it be possible for you to uncomment those lines (comment line 85) and try running again? I'm curious if we will see segfaults with the pi-pi stacking/cation-pi turned on.

lilleswing:
Updated feature set.

[full_grid.tar.gz](https://github.com/deepchem/deepchem/files/777411/full_grid.tar.gz)

lilleswing:
[core_grid.tar.gz](https://github.com/deepchem/deepchem/files/778588/core_grid.tar.gz)
[refined_grid.tar.gz](https://github.com/deepchem/deepchem/files/778589/refined_grid.tar.gz)

rbharath:
@lilleswing Something looks a little off here. The full grid is only 225 kB, while the old full grid is 5 mB. Could you check to see if the featurizer was failing silently under the hood? It will spit out empty vectors as features when that happens.

lilleswing:
blerg -- its attempting to calculate the full data-set here now.
http://171.65.103.54:8080/job/pdbbind-data-collection/5/console

We will see what it comes up with.

lilleswing:
Cached Value on S3 generated with python 2.7
```
shape=((11908, 2052), (11908, 1), (11908, 1), (11981,))
feature_vector_as_sets = 
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 34, 14, 16, 18, 20, 21, 22, 24])
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 16, 18, 20, 22, 26, 32, 40])
set([0, 1, 2, 3, 4, 6, 32, 8, 9, 10, 12, 14, 22, 26])
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 10, 12, 18])
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 10, 11, 12, 20, 23, 24])
non_zero count=4625177
```
New Values generated with python 3.5
```
shape=((11856, 2055), (11856, 1), (11856, 1), (11856,)) #correct it has fewer examples, but more features
feature_vector_as_sets = 
{0, 10, 6}
{0}
{0, 2}
{0, 2, 6}
{0, 4, 6}
non_zero_count=22386
```

However when I generate the first couple pdb with python 2.7...
```
shape=((12, 2055), (12, 1), (12, 1), (11856,))
feature_vector_as_sets=
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 34, 14, 16, 20, 22, 24])
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 14, 16, 40, 18, 19, 20, 22, 26, 30])
set([0, 1, 2, 3, 4, 6, 32, 8, 9, 10, 12, 14, 16, 22, 26, 28])
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 10, 12, 14])
set([0, 1, 2, 3, 4, 5, 6, 7, 8, 10, 12, 14, 15, 24, 26])
```

This seems to be an issue with not being able to properly calculate the OTHER features with python 3.5.  I'm starting a calculation using python 2.7 now.  Something to note is that the pickle files created by this process are not transferable between python 27 and python 35

rbharath:
@lilleswing Ah, good find. I think we should store these as 2.7 pickles for now since most of the deepchem developers use 2.7, but we need to start storing both 2.7 and 3.5/3.6 versions so we have it available for all users.

lilleswing:
[core_grid_27.tar.gz](https://github.com/deepchem/deepchem/files/783969/core_grid_27.tar.gz)
[full_grid_27.tar.gz](https://github.com/deepchem/deepchem/files/783970/full_grid_27.tar.gz)
[refined_grid_27.tar.gz](https://github.com/deepchem/deepchem/files/783971/refined_grid_27.tar.gz)

Here are the featurized datasets using python 2.7.  I'm still investigating why we are getting different features with python 3.5.  It seems to be an issue with the splif voxel featurizer.

evanfeinberg:
I suspect it could be a difference in how py27 vs py35 in handling of
hashing or deterministic order for converting sets to lists?

lilleswing:
Looking deeper into it the issues come up when Hydrogenating the protein.  The xyz coordinates of proteins are very different between python 2.7 and python 3.5 after Hydrogenation.

For the first protein in the dataset there are 301 atoms which have moved over 1 angstrom between the xyz coordinates of python 2.7 and python 3.5.  

This is causing the pairwise distances between the ligand and the protein to be larger, and thus making there exist fewer features with "hits" for splif.

rbharath:
Summarizing our offline discussion: The pi-pi/cation-pi bonds now work but seem to be a little buggy performance-wise. @evanfeinberg and I will debug this in a separate thread. Going to go ahead and merge this in. Thanks for the careful debugging @lilleswing!

LGTM.

PR #436, opened 2017-03-06
Rdkit Grid Featurizer For Pdbbind benchmark

lilleswing:
Gets comparable results on the benchmark without using Openbabel.  We use Pdbfixer to clean up the inputs.

|Algorithm | Dataset | Train | Validate | 
| -- | -- | --| --|
| OB | Refined | 0.963 | 0.511 |
| OB | Full | 0.983 | 0.528 | 
| rdkit | Refined | .98 | .565 |
| rdkit | Full | .97 | .48 | 

### Notes:
I still have not implemented Hydrogen Bonds.  Currently I assume no two atoms create hydrogens bonds with each other.

_is_salt_bridge needs some test cases.  While we are supposedly calculating the same charges I worry that there are differences in units/magnitudes between rdkit charges and OB charges.  We use the exact same algorithm and magnitudes for attempting to see if there are salt bridges.

I use Pdbfixer for BOTH the ligands and proteins. ~~In order to do the ligands I write the ligand to a temprary file in .pdb format so Pdbfixer can read it in.  It would be nice if Pdbfixer could read from a StringIO object instead of a string filename~~

You are supposed to use an encoding magic comment if you have utf-8 characters in your comments
https://www.python.org/dev/peps/pep-0263/

### Next Steps:
There is still a lot of usage of OpenBabel in the docking package (even though none of that code is used for any benchmarks).  I will slowly port it over attempting to keep passing existing tests.


peastman:
> It would be nice if Pdbfixer could read from a StringIO object instead of a string filename.

Use the `pdbfile` argument.  It will accept any file-like object.

deepchem/utils/rdkit_util.py
```diff
@@ -0,0 +1,117 @@
...
+  return (xyz)
+
+
+def add_hydrogens_to_mol(mol):
+  molecule_file = None
+  try:
+    molecule_file = str(tempfile.NamedTemporaryFile().name)
+    Chem.MolToPDBFile(mol, molecule_file)
+    fixer = PDBFixer(filename=molecule_file)
+    fixer.addMissingHydrogens(7.4)
```
proteneer: does this add Hs to the ligand as well? IIRC this is probably using something like a pKa table lookup for amino acids (and probably doesn't do anything for the ligand Hs)

deepchem/utils/rdkit_util.py
```diff
@@ -0,0 +1,117 @@
...
+  if my_mol is None:
+    raise ValueError("Unable to read non None Molecule Object")
+
+  if calc_charges:
+    my_mol = add_hydrogens_to_mol(my_mol)
+    compute_charges(my_mol)
+  elif add_hydrogens:
+    my_mol = add_hydrogens_to_mol(my_mol)
+    compute_charges(my_mol)
+
```
proteneer: what happens if you run rdmolops.SanitizeMol() after using PDBFixer to add the Hs? Do you still get errors?

proteneer: PS if you're going to be running a fingerprinter on this you probably most definitely want to sanitize this, since the finger prints generated are likely going to depend on perceived properties like Aromaticity, etc.

deepchem/feat/rdkit_grid_featurizer.py
```diff
@@ -0,0 +1,1266 @@
...
+      continue
+    env = Chem.FindAtomEnvironmentOfRadiusN(mol, degree, i, useHs=True)
+    submol = Chem.PathToSubmol(mol, env)
+    smile = Chem.MolToSmiles(submol)
+    ecfp_dict[i] = "%s,%s" % (mol.GetAtoms()[i].GetAtomicNum(), smile)
+
+  return ecfp_dict
+
+
+def _compute_ecfp_features(system_ob, ecfp_degree, ecfp_power):
```
rbharath: Can you change `system_ob` to an RDKit name?

deepchem/feat/rdkit_grid_featurizer.py
```diff
@@ -0,0 +1,1266 @@
...
+                                   pairwise_distances=None,
+                                   cutoff=4.5,
+                                   ecfp_degree=2):
+  """Computes ECFP dicts for ligand and binding pocket of the protein.
+
+  Parameters
+  ----------
+  protein_xyz: np.ndarray
+    Of shape (N_protein_atoms, 3)
+  protein: PDB object (TODO(rbharath): Correct?)
```
rbharath: Yes, but this is the PDB object from `nnscore_utils.py` (IIRC). Would be cool to see if we can get this to an rdkit/mdtraj object instead.

deepchem/utils/rdkit_util.py
```diff
@@ -0,0 +1,125 @@
...
+def add_hydrogens_to_mol(mol):
+  molecule_file = None
+  try:
+    pdbblock = Chem.MolToPDBBlock(mol)
+    pdb_stringio = StringIO()
+    pdb_stringio.write(pdbblock)
+    pdb_stringio.seek(0)
+    fixer = PDBFixer(pdbfile=pdb_stringio)
+    fixer.addMissingHydrogens(7.4)
```
proteneer: The input mol here is a structure that was generated with removeHs=False, so presumably it has some hydrogens already at some possibly non-physiological pH (eg. 9.0).

But then here we add some additionalHs at a different physiological pH (7.4) ? If so, wouldn't we want to remove all the Hs initially as well (the ones at 9.0)? Shouldn't you have the protonation states of the protein residues be consistent at a given pH?

deepchem/feat/rdkit_grid_featurizer.py
```diff
@@ -0,0 +1,1266 @@
...
+                                    ligand,
+                                    pairwise_distances=None,
+                                    cutoff=7.0):
+  """Computes Sybyl dicts for ligand and binding pocket of the protein.
+
+  Parameters
+  ----------
+  protein_xyz: np.ndarray
+    Of shape (N_protein_atoms, 3)
+  protein: PDB object (TODO(rbharath): Correct?)
```
rbharath: Same comment as above.

deepchem/utils/rdkit_util.py
```diff
@@ -0,0 +1,125 @@
...
+
```
proteneer: also, do you want to use PDBFixer to run a subsequent LocalEnergyMinimizer? @peastman may be better able to answer this

deepchem/feat/rdkit_grid_featurizer.py
```diff
@@ -0,0 +1,1266 @@
...
+
+
+def _merge_two_dicts(x, y):
+  """Given two dicts, merge them into a new dict as a shallow copy."""
+  z = x.copy()
+  z.update(y)
+  return z
+
+
+def _compute_centroid(coordinates):
```
rbharath: As a high level comment, this file has a number of functions of type `_name_of_function` for legacy reasons. Let's just go ahead and get rid of the first underscore to make them `name_of_function` (so `compute_centroid` here). A lot of these functions are general enough that we might as well expose them to users.

deepchem/feat/rdkit_grid_featurizer.py
```diff
@@ -0,0 +1,1266 @@
...
+  ligand_pi_t = {}
+
+  for protein_ring in ob.OBMolRingIter(protein):
+    if protein_ring.IsAromatic():
+      protein_ring_center = _compute_ring_center(protein, protein_ring)
+      protein_ring_normal = _compute_ring_normal(protein, protein_ring)
+
+      ligand_pi_parallel_idxs = set()
+      ligand_pi_t_idxs = set()
+      for ligand_ring in ob.OBMolRingIter(ligand):
```
rbharath: Ditto here with openbabel call.

deepchem/feat/rdkit_grid_featurizer.py
```diff
@@ -0,0 +1,1266 @@
...
+
+  return False
+
+
+def _compute_cation_pi(protein, ligand, protein_cation_pi, ligand_cation_pi):
+  for protein_ring in ob.OBMolRingIter(protein):
+    if protein_ring.IsAromatic():
+      protein_ring_center = _compute_ring_center(protein, protein_ring)
+      protein_ring_normal = _compute_ring_normal(protein, protein_ring)
+      for atom in ob.OBMolAtomIter(ligand):
```
rbharath: Ditto

rbharath:
@lilleswing This is awesome! Thanks for putting this PR in. I think that using PDBFixer to handle protein/ligand cleaning is a much better solution than our old Openbabel approach. I've made a couple more review comments here.

proteneer:
@lilleswing the latest rdkit can now successfully read 1jld with sanitize=True and removeHs=False, and it's very fast to boot.

lilleswing:
[full_grid.tar.gz](https://github.com/deepchem/deepchem/files/822732/full_grid.tar.gz)
[refined_grid.tar.gz](https://github.com/deepchem/deepchem/files/822733/refined_grid.tar.gz)

lilleswing:
[core_grid.tar.gz](https://github.com/deepchem/deepchem/files/822754/core_grid.tar.gz)

rbharath:
@lilleswing Is this one ready to merge? We might want to comment out the old `GridFeaturizer` to indicate people shouldn't use it.

evanfeinberg:
Awesome, thanks for all of your great efforts on this Karl! I've really
enjoyed starting to use RDKit lately and I'm glad it enables us to gain
easy install-ability and usability without any sacrifice (if anything, the
opposite) in performance.

cxhernandez:
> I still have not implemented Hydrogen Bonds. Currently I assume no two atoms create hydrogens bonds with each other.

MDTraj could helpful for that: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/hbond.py

evanfeinberg:
+1 to @cxhernandez

evanfeinberg:
with the caveat that loading individual molecules into ```mdtraj``` is
still relatively slow and will overall slow down featurization.
```mdtraj``` is quite fast for trajectories since compared to the startup
cost of loading topologies, loading each marginal frame after that is way
faster.

cxhernandez:
> with the caveat that loading individual molecules into ```mdtraj``` is still relatively slow and will overall slow down featurization.

If you're already loading the system into PDBFixer, you should be able to create an `mdtaj.Trajectory` object without much trouble and shouldn't have to reload the trajectory.

evanfeinberg:
awesome, since it's no added time cost then I definitely advocate trying
out mdtraj hbonds.

lilleswing:
Closed in favor of
https://github.com/deepchem/deepchem/pull/440

PR #511, opened 2017-04-11
Small updates for atomic convs

lilleswing:
Test Dataset can't be transformed as it doesn't have y values -- I changed it to a prediction.

Main change is the documentation of the atomic convolution layer getting the number of channels correct.

This is surprising to me -- my intuition in line with convolutions would be to SUM over all the atom type masks instead of making each one a feature.

### DIFF
--- a/contrib/atomicconv/acnn/core/opt_scaffold.py
+++ b/contrib/atomicconv/acnn/core/opt_scaffold.py
@@ -40,6 +40,7 @@ transformers = [
 ]
 for transformer in transformers:
   train_dataset = transformer.transform(train_dataset)
+  test_dataset = transformer.transform(test_dataset)
 
 at = [1., 6, 7., 8., 9., 11., 12., 15., 16., 17., 20., 25., 30., 35., 53.]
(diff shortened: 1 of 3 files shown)


joegomes:
@lilleswing The test dataset transformer error was caused by an empty `tasks` attribute in the dataset.  I've re-uploaded the fixed dataset, can you try evaluating again?

Summing over atom types will throw out any information you give the network about atom types.  We sum over all neighbor atoms of a specific type for a given radial filter.

rbharath:
@lilleswing Changes look good with only yapf failures. Let me know once you think it's ready to merge.

I'd recommend trying a head-to-head test with both variants. There are many architectural choices which can be tweaked here.

joegomes:
I agree it would be good to compare.

lilleswing:
Well my intuition is wrong until we start learning atom_type filters instead of one-hot encoding for each atom type..  Pushed update which should fix travis and reverted opt_scaffold as it works now.

rbharath:
LGTM.

PR #568, opened 2017-05-22
DeepChem learns to play tictactoe

lilleswing:
I played around with a3c and RL in Deepchem seeing if I could train a tictactoe agent.

The agent isn't good yet and is still having issues.

1) after training for a long time the Dense networks output NaN instead of float32 values
2) The agent just makes illegal moves constantly.

I still wanted to put this  up for CR because I found some small things in A3C that might trip up others so I wanted to fix them.

1) Deep Copy the environment's state in the worker thread.  Environments will most likely mutate their environment on step commands, without this all states would just be the final state for the rollout.

2) Ininitalize the global variables after the worker threads are created.  This might be wrong @peastman .  This was needed because workers were creating their own variables on worker thread creation.  How did you get it to work otherwise?  Do we want the layers shared between the workers or each worker to have their own network?

### DIFF
--- a/deepchem/rl/a3c.py
+++ b/deepchem/rl/a3c.py
@@ -21,11 +21,10 @@ class A3CLoss(Layer):
     self.entropy_weight = entropy_weight
 
   def create_tensor(self, **kwargs):
-    reward, action, prob, value, advantages = [
-        layer.out_tensor for layer in self.in_layers
-    ]
+    reward, action, prob, value = [layer.out_tensor for layer in self.in_layers]
     log_prob = tf.log(prob + 0.0001)
-    policy_loss = -tf.reduce_sum(advantages * tf.reduce_sum(action * log_prob))
+    policy_loss = -tf.reduce_sum(
+        (reward - value) * tf.reduce_sum(action * log_prob))
     value_loss = tf.reduce_sum(tf.square(reward - value))
     entropy = -tf.reduce_sum(prob * log_prob)
     self.out_tensor = policy_loss + self.value_weight * value_loss - self.entropy_weight * entropy
@@ -91,8 +90,7 @@ class A3C(object):
     self.optimizer = TFWrapper(
         tf.train.AdamOptimizer, learning_rate=0.001, beta1=0.9, beta2=0.999)
     (self._graph, self._features, rewards, actions, self._action_prob,
-     self._value,
-     self._advantages) = self._build_graph(None, 'global', model_dir)
+     self._value) = self._build_graph(None, 'global', model_dir)
     with self._graph._get_tf("Graph").as_default():
       self._session = tf.Session()
 
@@ -103,12 +101,11 @@ class A3C(object):
     action_prob = policy_layers['action_prob']
     value = policy_layers['value']
     rewards = Weights(shape=(None, 1))
-    advantages = Weights(shape=(None, 1))
     actions = Label(shape=(None, self._env.n_actions))
     loss = A3CLoss(
         self.value_weight,
         self.entropy_weight,
-        in_layers=[rewards, actions, action_prob, value, advantages])
+        in_layers=[rewards, actions, action_prob, value])
     graph = TensorGraph(
         batch_size=self.max_rollout_length,
         use_queue=False,
@@ -123,7 +120,7 @@ class A3C(object):
     with graph._get_tf("Graph").as_default():
       with tf.variable_scope(scope):
         graph.build()
-    return graph, features, rewards, actions, action_prob, value, advantages
+    return graph, features, rewards, actions, action_prob, value
 
   def fit(self, total_steps, max_checkpoints_to_keep=5,
           checkpoint_interval=600):
@@ -227,7 +224,7 @@ class _Worker(object):
     self.scope = 'worker%d' % index
     self.env = copy.deepcopy(a3c._env)
     self.env.reset()
-    self.graph, self.features, self.rewards, self.actions, self.action_prob, self.value, self.advantages = a3c._build_graph(
+    self.graph, self.features, self.rewards, self.actions, self.action_prob, self.value = a3c._build_graph(
         a3c._graph._get_tf('Graph'), self.scope, None)
     with a3c._graph._get_tf("Graph").as_default():
       local_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
@@ -246,14 +243,12 @@ class _Worker(object):
       session = self.a3c._session
       while step_count[0] < total_steps:
         session.run(self.update_local_variables)
-        episode_states, episode_actions, episode_rewards, episode_advantages = self.create_rollout(
-        )
+        episode_states, episode_actions, episode_rewards = self.create_rollout()
         feed_dict = {}
         for f, s in zip(self.features, episode_states):
           feed_dict[f.out_tensor] = s
         feed_dict[self.rewards.out_tensor] = episode_rewards
         feed_dict[self.actions.out_tensor] = episode_actions
-        feed_dict[self.advantages.out_tensor] = episode_advantages
         session.run(self.train_op, feed_dict=feed_dict)
         step_count[0] += len(episode_actions)
 
@@ -264,7 +259,6 @@ class _Worker(object):
     states = [[] for i in range(len(self.features))]
     actions = []
     rewards = []
-    values = []
     for i in range(self.a3c.max_rollout_length):
       if self.env.terminated:
         break
@@ -272,29 +266,20 @@ class _Worker(object):
       for j in range(len(state)):
         states[j].append(state[j])
       feed_dict = _create_feed_dict(self.features, state)
-      probabilities, value = session.run(
-          [self.action_prob.out_tensor, self.value.out_tensor],
-          feed_dict=feed_dict)
+      probabilities = session.run(
+          self.action_prob.out_tensor, feed_dict=feed_dict)
       action = np.random.choice(np.arange(n_actions), p=probabilities[0])
       actions.append(np.zeros(n_actions))
       actions[i][action] = 1.0
-      values.append(value)
       rewards.append(self.env.step(action))
     if not self.env.terminated:
       # Add an estimate of the reward for the rest of the episode.
       feed_dict = _create_feed_dict(self.features, self.env.state)
       rewards[-1] += self.a3c.discount_factor * session.run(
           self.value.out_tensor, feed_dict)
-    gamma = 0.99
-    advantages = []
-    for i in range(1, len(rewards)):
-      advantages.append(rewards[i - 1] + gamma * (values[i] - values[i - 1]))
-    advantages.append(rewards[-1] - values[-1])
-
     for j in range(len(rewards) - 1, 0, -1):
       rewards[j - 1] += self.a3c.discount_factor * rewards[j]
-
     if self.env.terminated:
       self.env.reset()
     return np.array(states), np.array(actions), np.array(rewards).reshape(
-        (len(rewards), 1)), advantages
+        (len(rewards), 1))
--- a/deepchem/rl/a3c.py
+++ b/deepchem/rl/a3c.py
@@ -22,7 +22,8 @@ class A3CLoss(Layer):
 
   def create_tensor(self, **kwargs):
     reward, action, prob, value = [layer.out_tensor for layer in self.in_layers]
-    log_prob = tf.log(prob + 0.0001)
+    prob = prob + np.finfo(np.float32).eps
+    log_prob = tf.log(prob)
     policy_loss = -tf.reduce_sum(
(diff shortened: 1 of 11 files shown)


peastman:
Thanks, I'll take a look!

deepchem/rl/a3c.py
```diff
@@ -237,7 +237,7 @@ def __init__(self, a3c, index):
       self.train_op = a3c._graph._get_tf('Optimizer').apply_gradients(
           grads_and_vars)
       self.update_local_variables = tf.group(
-          * [tf.assign(v1, v2) for v1, v2 in zip(local_vars, global_vars)])
+          *[tf.assign(v1, v2) for v1, v2 in zip(local_vars, global_vars)])
```
peastman: yapf is complaining about this change.  It's one of those super irritating cases where yapf produces a different result on some computers than others, so if you run it locally, it produces formatting that travis rejects as wrong.

deepchem/rl/a3c.py
```diff
@@ -263,7 +263,7 @@ def create_rollout(self):
     for i in range(self.a3c.max_rollout_length):
       if self.env.terminated:
         break
-      state = self.env.state
+      state = copy.deepcopy(self.env.state)
```
peastman: This shouldn't be needed.  States should be immutable.  The problem is that you've implemented your environment so that it modifies states after outside code has already retrieved them.  Doing that opens the door to all sorts of bugs in lots of places.  Instead it should assign a new value to `_state` which is then treated as immutable.

peastman:
> How did you get it to work otherwise?

No idea!  Your change looks correct though.

> Do we want the layers shared between the workers or each worker to have their own network?

Each worker needs its own network.  This is actually one of the things that makes A3C more robust: different workers may be using slightly different networks at the same time, which increases the diversity of trajectories they explore.

deepchem/rl/a3c.py
lilleswing: While I understand that decision -- I don't think it will be clear to most users and will lead to users mutating state in the Environment with step() and reset().  We should at least document it on the Environment object, state property, and step function.

Doing it here is safer and easier to reason about -- but will lead to worse performance in the case of someone actually making their state immutable, and creating a new one each time.

lilleswing: I can change my tictactoe board and document the functions though if we decide to ask for immutability of state by the env.

peastman: We should definitely document it for people writing environments.  It's not just this one line—it would create opportunities for bugs all over the place.  Any piece of code that ever stores a state would start having to worry that state might magically change behind its back.

lilleswing:
The NaNs were from tf.log(prob) where prob was a probability vector.  A single value which == tf.float32(0) would destroy the whole network.  I added an epsilon to get around this.

rbharath:
I really like the idea of adding more environments to DeepChem. New environments are the only real way of benchmarking our RL suite, and it would be great to build up our capabilities. In fact, I have a couple extra suggestions:

- Can you make a new submodule `dc.rl.envs` that hosts environments and move the Tic-tac-toe environment there?
- Can you add a section to the benchmarks on the `README.md` that documents best performance on the Tic-Tac-Toe environment? I think we should start keeping running numbers on the RL environments like we do for MoleculeNet (betting RL is going to be increasingly important moving forward).

peastman:
If you want lots of environments, you could just make OpenAI Gym an optional dependency.  We've already got support for it via the GymEnvironment class, so no new code is even required.  Just start adding benchmarks or examples that use it.

rbharath:
Definitely like the idea of making OpenAI gym an optional dependency. Any PRs adding benchmarks or examples in this vein welcome.

deepchem/rl/a3c.py
```diff
@@ -21,12 +21,13 @@ def __init__(self, value_weight, entropy_weight, **kwargs):
     self.entropy_weight = entropy_weight
 
   def create_tensor(self, **kwargs):
-    reward, action, prob, value = [layer.out_tensor for layer in self.in_layers]
-    log_prob = tf.log(prob)
+    reward, action, prob, value, advantages = [layer.out_tensor for layer in self.in_layers]
+    log_prob = tf.log(prob + 0.0001)
     policy_loss = -tf.reduce_sum(
```
lilleswing: So isn't this just another value loss?  What is encouraging the network to seed out better rewards/values?  

The way I am reading this loss function right now is it is encouraging certainty -- that is the value loss, policy loss both encourage the value==discounted_rewards.  The entropy also encourages certainty, to not be unsure but to go where you know where you have accurate value function.  

I implemented advantage as seen here
https://medium.com/emergent-future/simple-reinforcement-learning-with-tensorflow-part-8-asynchronous-actor-critic-agents-a3c-c88f72a5e9f2
In order to make the policy loss portion encourage going towards higher rewards.

Still not able to play tic tac toe though.

Am I reading the current implementation of policy loss wrong @peastman ?

deepchem/rl/a3c.py
```diff
@@ -21,10 +21,11 @@ def __init__(self, value_weight, entropy_weight, **kwargs):
...
   def create_tensor(self, **kwargs):
-    reward, action, prob, value = [layer.out_tensor for layer in self.in_layers]
-    log_prob = tf.log(prob)
-    policy_loss = -tf.reduce_sum(
-        (reward - value) * tf.reduce_sum(action * log_prob))
+    reward, action, prob, value, advantages = [
+        layer.out_tensor for layer in self.in_layers
+    ]
+    log_prob = tf.log(prob + 0.0001)
+    policy_loss = -tf.reduce_sum(advantages * tf.reduce_sum(action * log_prob))
```
peastman: I don't understand what you're doing here.  Advantage is defined as reward-value, where the reward is the actual reward that was obtained in the rollout, and value is the value estimate produced by the policy.  That's precisely what it was already using.

lilleswing: I don't think that is the definition of advantage.
From the A3C paper
https://arxiv.org/pdf/1602.01783.pdf

selection_094

What I implemented here is k=1

lilleswing: Page 4 bottom left corner

lilleswing:
@rbharath what do we mean by performance?  

* avg score of a rollout?
  * How much do we allow to train?
* Time to a given avg score?
* Max score of a rollout?

Performance for these things is *hard* because everyone disagrees on what performance means.

rbharath:
On my phone, so will respond about perf in more depth later. In meantime,
this new effort by OpenAI is very pertinent

https://blog.openai.com/openai-baselines-dqn/amp/

deepchem/rl/a3c.py
peastman: That isn't necessarily the correct equation to use.  They say the update "can be seen as" an estimate of that function.  However, the actual loss function they use appears two paragraphs later.  It just depends on `R-V` without that extra term.

Also see https://arxiv.org/pdf/1506.02438.pdf for a much more detailed discussion of advantage estimation.  They present both forms (eq. 17 and 18) as limiting cases of a continuous family of advantage estimates (eq. 16).  The one I implemented (`lambda=1`) is unbiased but has larger variance.  The one you used (`lambda=0`) reduces the variance but introduces bias.  In their experiments, they get the best results for values in the range of about 0.96, so much closer to 1 than 0.  We could consider adding `lambda` as a user selectable hyperparameter.

lilleswing: Ok I'll update A3C to have lambda as a hyper parameter.  Without the change in advantage my agent does not learn to play tictactoe very well.

I'm interested what the grid of avg reward looks like for my network for differing lambda and gamma.  After I implement lambda as a hyper param I'll update with my grid of avg reward along with the code.

peastman: Sounds good.

peastman: I'd also suggest experimenting with different values for `value_weight`.  It's possible the problem is just that your weight is too low, so it's taking a long time to learn the value function.

Also, are you sure it's converged?  100,000 steps isn't very many.

peastman: Try going up to much larger values for `value_weight`.  A possible explanation for what you see is just that it isn't learning the value function (because you've set the weight too low).

peastman: You also could try directly querying the value function for various states, to see whether it's learned it or not.

lilleswing: @peastman you were right there was a value_weight sweat spot between 4-6.

selection_003

selection_004

The model isn't as stable as with lambda=0 but if you use a pocket algorithm it can produce similar results.

peastman: Good to know.  It will be interesting to see the full matrix of how it depends on gamma, lambda, and value_weight.

deepchem/rl/a3c.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, value_weight, entropy_weight, **kwargs):
 
   def create_tensor(self, **kwargs):
     reward, action, prob, value = [layer.out_tensor for layer in self.in_layers]
-    log_prob = tf.log(prob)
+    log_prob = tf.log(prob + 0.0001)
```
peastman: Why 0.0001?  If the goal is just to make sure we never take the log of zero, then machine epsilon seems like a better value.  You can get it as `np.finfo(np.float32).eps`.  Also, might it be better to use `max(prob, eps)` rather than addition?  That will distort the probability distribution less.

peastman:
Aside from the one comment I just made about the additive constant, this looks good to me.

peastman:
By the way, after looking at the equation you highlighted in the paper again I realized I was misinterpreting it.  The extra term is nothing more than adding a single V term to estimate everything beyond the end of the rollout.  We already include that.

https://github.com/deepchem/deepchem/blob/master/deepchem/rl/a3c.py#L276-L280

lilleswing:
@peastman huh I misread that chunk of code assuming it was doing some sort of simulation for the rest of a rollout if it ended early.

That makes the improved performance after my changes to advantage even more unexplained.

peastman:
It's only one term, not a sum of terms.  Remember, V is the *total* expected return starting from that state, not the expected return from just one step.  If the task has terminated, it's zero by definition (there will be no more rewards), so there's nothing to add.  But if you ended the task early because you reached the maximum rollout length, you want to add an estimate of the additional reward you would have gotten if you hadn't stopped there.

lilleswing:
I'm going to summarize what I learned here for the github issue.  I didn't realize code lines that were overwritten causes the comments to get lost.

When lambda=0 for the given state you explicitly add the difference in value function between the two consecutive states as part of the policy loss.  This is effectively leaning as hard as you can on short term gain.

When lambda=1 this is favoring long term gain.  A "bad move" at the very end of your game gets discounted through the whole rollout and used as the "score" of every policy step.

I suspect why lambda=0 works for tictactoe is the main issue is making illegal moves late in the game.  An illegal move late in the game does not necessarily mean every move before has been bad -- but the discounted reward for each state up to that point will be a poor score.

https://arxiv.org/pdf/1506.02438.pdf is a paper which goes more in depth about Generalized Advantage Estimation (Michael Jordan and Pieter Abbeel).

deepchem/rl/a3c.py
lilleswing: I added np.finfo(np.float32).eps.  I'm willing to take the hit into the distortion of probability distribution for the added simplicity.

peastman:
Not quite.  The difference isn't a matter of favoring short term vs. long term gain.  (That's what gamma is for.)  The difference is how you estimate long term gain.  There are two ways you might try to estimate the value function given the state at step i: `V(s_i) = r_i + r_(i+1) + r_(i+2) + ...` or `V(s_i) = r_i + V(s_(i+1))`.  (For simplicity I'm assuming gamma is 1.)  The first one is unbiased, but has a high variance since it's a sum of many terms.  The second one is much lower variance, since it only depends on one step, but it assumes you already know how to calculate the value function for the next step.  But that's the quantity we're trying to learn, so all we can do is use our current estimate of it, and that introduces bias.  The purpose of lambda is to interpolate between these two approximations.

As for why you get better results with lambda=0, I'd say it's because RL is super finicky and tiny changes can have big effects on the results, often not for any obvious reason. :)

rbharath:
Looked through the code and things look good to me. The TicTacToe seems like a simple, but quite nontrivial learning problem, so will be a good test bed. Looking forward to seeing how DQN style methods do on it.

LGTM.

rbharath:
Apparently pull requests with a failing status can't be merged on a phone... Will merge in once back at desk.

rbharath:
Looks like there's a yapf failure. Good to merge in once that's fixed.

lilleswing:
@rbharath @peastman sorry I was on vacation and out of internet contact for the past 1.5 weeks.  I fixed the YAPF problem and this is good to go now.

rbharath:
@lilleswing Awesome, thanks for the fix. LGTM.

PR #607, opened 2017-06-21
Install tensorflow with conda

lilleswing:
Install Tensorflow with Conda.

I'll also test after Travis passes if we can upgrade to 1.1.0.

### DIFF
--- a/scripts/install_deepchem_conda.sh
+++ b/scripts/install_deepchem_conda.sh
@@ -19,9 +19,9 @@ fi
 
 if [ "$python_version" == "3.5" ]
 then
-    export protobuf_url=https://storage.googleapis.com/tensorflow/linux/cpu/protobuf-3.1.0-cp35-none-linux_x86_64.whl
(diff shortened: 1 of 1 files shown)


lilleswing:
Installing tensorflow-gpu via conda causes rdkit to roll back to it's 2015 version which breaks a lot of things :(

selection_016

mwojcikowski:
Do you have omnia channel active all the time? I had similar issues with rdkit, for some reason conda packages of RDKit lock Numpy version, but RDKit itself is really generous in that term... 

Maybe it would be best to separate deepchem channel with no further dependencies, or at least fewer like conda-forge.

lilleswing:
@mwojcikowski a single conda install would be great.  We have omnia deepchem but it is very old (< 1.0.0).  Any work on a single conda recipe would be greatly appreciated!  (Right now we are running into more installation problems due to dependencies changing and lib_boost even on mainline)

rbharath:
@lilleswing I believe @cxhernandez updated omnia to run the latest DeepChem recently.

PR #638, opened 2017-06-30
Account for y having a channel in normalizing transformer

lilleswing:
Account For Y having a channel in normalizing transformer.

When pulling an output from TensorGraph we add a channel even if it is of size 1.  We have to account for this in the y transformer untransform.


lilleswing:
I split this into a seperate CR as it would be more controversial than save/load weave layers.

lilleswing:
This would break things if y was multi-dimensional (say having images).  But its currently broken now...

PR #724, opened 2017-08-04
Removing comment that pcba_datasets doesnt do train/test split as I b…

LRParser:
…elieve it's incorrect


lilleswing:
Yep -- it does do the split.  You need to yapf

```bash
pip install yapf==0.16.2
yapf -i examples/pcba/pcba_datasets.py
```

LRParser:
Will do thanks

LRParser:
I should have rebased - will try this as part of a larger PR.

PR #855, opened 2017-09-27
Changes to support GANs

### LINKED ISSUE #535: Adding GAN support to TensorGraph

peastman:
See #535.  This turned out to be a bit more involved than I thought, but I think all the changes are reasonable.  For example, it now runs variable initializers in `build()` rather than `fit()`, since reinitializing them every time you call `fit()` obviously wouldn't make sense for this case.  Next step will be to create an example that trains a GAN.  That should give us a better sense of how this API works in practice.

There's one test failure, which is an actual bug that was exposed by these changes.  `WeaveLayer` is setting its `out_tensor` field to a list instead of a tensor.  Someone else needs to fix that.


peastman:
The Python 3 build is failing with an unrelated error:

```
ERROR: Failure: ImportError (libicui18n.so.56: cannot open shared object file: No such file or directory)
```

Could this be related to the upgrade to Tensorflow 1.3?

Dgelemi:
Which version of cuDNN are you using? Tensorflow 1.3 need version 6.
Not sure if it’s related though.


On 27 Sep 2017, at 19:49, peastman <notifications@github.com> wrote:

The Python 3 build is failing with an unrelated error:

ERROR: Failure: ImportError (libicui18n.so.56: cannot open shared
object file: No such file or directory)

Could this be related to the upgrade to Tensorflow 1.3?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/deepchem/deepchem/pull/855#issuecomment-332619525>, or mute
the thread

peastman:
It's not using CUDA.  Travis doesn't have GPUs.

lilleswing:
@peastman it was an issue importing rdkit not tensorflow -- no less confusing though.

peastman:
Ok.  Are you fixing it?

lilleswing:
I reran the build to see if it was reproducible -- if it is I can look at it at 3:00pm PST.

peastman:
I already reran it once.  It happened both times.

lilleswing:
Yeah looks like anaconda updated its version of icu and rdkit didn't pin it
https://anaconda.org/anaconda/icu
https://github.com/conda-forge/matplotlib-feedstock/issues/36

peastman:
Great, thanks!

proteneer:
pinging @greglandrum on this regarding what @lilleswing - should rdkit pin it?

deepchem/models/tensorgraph/tensor_graph.py
```diff
@@ -756,3 +838,40 @@ def __init__(self, tf_class, **kwargs):
 
   def __call__(self):
     return self.tf_class(**self.kwargs)
+
+
+class Submodel(object):
```
rbharath: Should a submodel be a top level class or perhaps inherit from one of our existing abstract classes?

rbharath:
@peastman Would you mind raising a separate issue for the `WeaveLayer` failure that was exposed with a minimum failing example? We can fix that on a separate thread.

deepchem/models/tensorgraph/tensor_graph.py
peastman: Which one did you have in mind?

rbharath: It's actually a little tricky now that I think about it. It feels like `Submodel` and `Model` should be related somehow but maybe that just doesn't work out under Python's current inheritance semantics.

peastman: It's nothing to do with Python.  They're just different concepts.  Remember I suggested "submodel" should really be called "training objective" instead?  That's what it really is.  You preferred the name "submodel", so that's what I used, but remember that it *isn't* a model.  It's a training objective.

rbharath: Let's rename it in that case :). I was envisioning something more related to a model object itself.

greglandrum:
@proteneer, @lilleswing : we don't specify where ICU comes from and use whatever's on the system.

I can take a look at pinning a particular version based on the boost version we pin.

rbharath:
@peastman Spent a bit more time thinking. I think we might simply be able to remove submodels/training-objectives and just allow a model to have multiple losses. Then a GAN is simply a model with two losses (discriminator loss and generator loss). Does this simplification work?

peastman:
The submodel/objective can specify a list of layers, a loss, and an optimizer.  Just one of those isn't enough.  For example, the generator's loss function depends on the discriminator's variables, but it shouldn't modify those when optimizing the generator.

rbharath:
@peastman This cgan tutorial looks amazing! Only thing I'd request is adding a paragraph describing how CGANs differ from regular GANs

peastman:
> Only thing I'd request is adding a paragraph describing how CGANs differ from regular GANs

It already does: "A Conditional GAN (CGAN) allows additional inputs to the generator and discriminator that their output is conditioned on. For example, this might be a class label, and the GAN tries to learn how the data distribution varies between classes."

peastman:
I believe this is now ready to merge.  `test_Weave_pickle()` is still failing, but that's a real error so it *should* be failing.  It will be fixed by #860.

rbharath:
@peastman Everything looks to pass. Congrats on a major feature add!

LGTM

PR #968, opened 2017-12-13
Make sure to run all doctests

lilleswing:


lilleswing:
This adds to our already long build time for something that I don't see adding to much value so I'm going to drop this PR

PR #1004, opened 2018-01-03
Create SubModule after loading TensorGraph From Disk

lilleswing:
This builds on top of the tensorflow 1.4 build.

Adding a submodule after loading from disk can be used for lots of things e.x.
1) Transfer learning from a very large ML run over large dataset
2) Fine tuning of the loss function for very large runs


deepchem/models/tensorgraph/models/test_graph_models.py
```diff
@@ -79,3 +80,22 @@ def test_graph_conv_error_bars(self):
         dataset, transformers, untransform=True, n_passes=24)
     assert mu.shape == (len(dataset), len(tasks))
     assert sigma.shape == (len(dataset), len(tasks))
+
+  def test_change_loss_function(self):
```
lilleswing: Example of fine tuning/transfer learning after loading in a saved model.

deepchem/models/tensorgraph/tensor_graph.py
```diff
@@ -788,7 +784,13 @@ def restore(self, checkpoint=None):
...
       raise ValueError('No checkpoint found')
     with self._get_tf("Graph").as_default():
-      saver = tf.train.Saver()
+      reader = NewCheckpointReader(checkpoint)
+      var_names = set([x for x in reader.get_variable_to_shape_map()])
+      var_map = {
+          x.op.name: x
+          for x in tf.global_variables() if x.op.name in var_names
+      }
+      saver = tf.train.Saver(var_list=var_map)
```
lilleswing: If you don't pass in a var map saver.restore throws an exception of your Graph doesn't have the exact same variables as in the checkpoint.  This new restore restores variables if they exist in the checkpoint -- otherwise it leaves the default initializer values from build()

lilleswing:
Made a clean one off of master -- 
https://github.com/deepchem/deepchem/pull/1007

PR #1012, opened 2018-01-05
Pypi upload

lilleswing:
Upload to pypi on successful travis builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,15 @@ install:
 - pip install coveralls
 - python setup.py install
 script:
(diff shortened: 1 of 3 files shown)


rbharath:
@lilleswing Awesome! Does this get us pip installs?

LGTM

lilleswing:
Pip installs without dependencies

rbharath:
Cool. So `pip install deepchem` now works?

lilleswing:
I believe that command will install the latest successful master build
without any of the dependencies.  (I have to double check though what the
default no arg version of pip install is)

lilleswing:
You could check via —

‘’’ bash
pip install deepchem
pip freeze | grep deepchem
‘’’

PR #1026, opened 2018-01-17
[READY] Variational Autoencoder Asupuru-Guzik

lilleswing:
WIP until I solve the issue of exploding gradients I am currently seeing.


lilleswing:
The Final Model

selection_033

There are some minor differences with the model from the paper.  All stated in the docstring.
The annealing function for the kl divergence is difference and there is no teacher forcing.  I also added a BatchNorm before the embedding layer.

deepchem/models/tensorgraph/layers.py
```diff
@@ -424,18 +469,21 @@ def create_tensor(self, in_layers=None, set_tensors=True, **kwargs):
...
-        [self.width, parent_channel_size, self.out_channels]))
-    t = tf.nn.conv1d(parent, f, stride=self.stride, padding=self.padding)
-    if self.biases_initializer is not None:
-      b = tf.Variable(self.biases_initializer()([self.out_channels]))
-      t = tf.nn.bias_add(t, b)
-    if self.activation_fn is None:
-      out_tensor = t
-    else:
-      out_tensor = self.activation_fn(t)
+    out_tensor = tf.keras.layers.Conv1D(
```
rbharath: Nice swap for keras here! We should make these swaps whenever we get the chance.

deepchem/models/tensorgraph/tests/test_seqtoseq.py
```diff
@@ -58,6 +58,52 @@ def test_int_sequence(self):
...
+
+    s.fit_sequences(generate_sequences(train_smiles, 100))
+
+    # Test it out.
+    pred1 = s.predict_from_sequences(train_smiles, beam_width=1)
+    pred4 = s.predict_from_sequences(train_smiles, beam_width=4)
+    embeddings = s.predict_embeddings(train_smiles)
+    pred1e = s.predict_from_embeddings(embeddings, beam_width=1)
+    pred4e = s.predict_from_embeddings(embeddings, beam_width=4)
+    count1 = 0
```
rbharath: It looks like `count1` and `count4` aren't used in the asserts. Should probably be removed.

lilleswing: Yeah they should -- I tried to get it to overfit a single string but wasn't able to figure out how to get it in a reasonable amount of time.

rbharath:
LGTM

Issue #1030, opened 2018-01-20
missing tasks when merging DiskDataset


fmonta:
Hello,
I have come up with the problem that, after merging pre-defined cross-validation folds using DiskDataset.merge(), the resulting dataset does not contain any tasks anymore, hence cannot be transformed (because get_shape() returns empty tuples for y and w). I believe a fix would be to edit the return of the merge() method
   `    return DiskDataset.create_dataset(generator(), data_dir=merge_dir)`
 to include somehow the tasks that were included in the individual datasets we are merging.

lilleswing:
Fixed with #1031

PR #1031, opened 2018-01-21
1030 - Tasks in dc.data.DiskDataset.merge

fmonta:
This PR is a fix for #1030, however that fix assumes that the DiskDatasets to merge have the same number of tasks. This works for my use case but not sure about the general contract of that method.
I also added a unittest for merge().

### DIFF
--- a/deepchem/data/datasets.py
+++ b/deepchem/data/datasets.py
@@ -915,12 +915,27 @@ class DiskDataset(Dataset):
     else:
       merge_dir = tempfile.mkdtemp()
 
+    # Protect against generator exhaustion
+    datasets = list(datasets)
+
+    # This ensures tasks are consistent for all datasets
+    tasks = []
+    for dataset in datasets:
+      try:
+        tasks.append(dataset.tasks)
+      except AttributeError:
+        pass
+    if tasks:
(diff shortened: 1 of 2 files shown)


deepchem/data/datasets.py
```diff
@@ -920,7 +920,7 @@ def generator():
         X, y, w, ids = (dataset.X, dataset.y, dataset.w, dataset.ids)
         yield (X, y, w, ids)
```
lilleswing: We need to do a check here that we are passing in only DiskDatasets.

```python
tasks = []
for dataset in datasets:
  if isintance(dataset, deepchem.data.DiskDataset):
    tasks = dataset.tasks
    break
return DiskDataset.create_dataset(generator(), data_dir=merge_dir, tasks=tasks)
```

lilleswing:
Great catch!  after the type check this should be good to go!

lilleswing:
just yapf left
```python
pip install yapf==0.20.0
yapf -i deepchem/data/datasets.py
deepchem/data/tests/test_datasets.py
```

fmonta:
done

lilleswing:
LGTM

PR #1032, opened 2018-01-22
Simplify logic in the tox21 load function.

tarokiritani:
This change makes the code a bit more readable.


lilleswing:
LGTM 

Congratulations on your first commit to DeepChem!

PR #1041, opened 2018-01-24
[WIP] Predict Loss as Output

lilleswing:
Allows predicting the loss as output
```python
import numpy as np
import deepchem as dc

# Only for debug!
np.random.seed(123)

# Load Tox21 dataset
n_features = 1024
tox21_tasks, tox21_datasets, transformers = dc.molnet.load_tox21()

# Fit models

model = dc.models.MultiTaskClassifier(
    len(tox21_tasks),
    n_features,
    layer_sizes=[1000],
    dropouts=[.25],
    learning_rate=0.001,
    batch_size=50,
    use_queue=False)

# Fit trained model
model.fit(train_dataset, nb_epoch=1)
valid_loss = model.predict(valid_dataset, outputs=model.loss)
print(np.mean(valid_loss))
```

Bug is brought up here https://github.com/deepchem/deepchem/issues/1040

### DIFF
--- a/deepchem/models/tensorgraph/tensor_graph.py
+++ b/deepchem/models/tensorgraph/tensor_graph.py
@@ -279,11 +279,11 @@ class TensorGraph(Model):
           deterministic=deterministic,
           pad_batches=pad_batches):
         feed_dict = dict()
-        if len(self.labels) == 1 and y_b is not None and not predict:
+        if len(self.labels) == 1 and y_b is not None:
           feed_dict[self.labels[0]] = y_b
         if len(self.features) == 1 and X_b is not None:
           feed_dict[self.features[0]] = X_b
-        if len(self.task_weights) == 1 and w_b is not None and not predict:
+        if len(self.task_weights) == 1 and w_b is not None:
           feed_dict[self.task_weights[0]] = w_b
         for (initial_state, zero_state) in zip(self.rnn_initial_states,
                                                self.rnn_zero_states):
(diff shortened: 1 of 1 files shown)


deepchem/models/tensorgraph/tensor_graph.py
```diff
@@ -279,11 +279,11 @@ def default_generator(self,
           deterministic=deterministic,
           pad_batches=pad_batches):
         feed_dict = dict()
-        if len(self.labels) == 1 and y_b is not None and not predict:
+        if len(self.labels) == 1 and y_b is not None:
```
lilleswing: So I initially added the predict flag in April.  
Back then we did not allow arbitrary outputs of predict.

We already have a None check for the generator.  Is there any reason besides performance to not always pass in the tensor?  If a user is a power user and wants performance they can create a dataset without None values

deepchem/models/tensorgraph/tensor_graph.py
```diff
@@ -349,7 +349,11 @@ def predict_on_generator(self, generator, transformers=[], outputs=None):
 
       final_results = []
       for result_list in results:
-        final_results.append(np.concatenate(result_list, axis=0))
+        try:
```
lilleswing: Since loss is an aggregated statistic it can't be concatenated (only a single entry per mini-batch).

We shouldn't fail hard in these cases -- but instead just return the results to the user chunked into mini-batches.

lilleswing:
Still need to add tests -- just wanted to see what the CI system says.

fuzunyun:
default
default

fuzunyun:
I tried your code but some error occured which I have no idea how it happened. Could you please help me find out where the problem is?  Attached is my full script.
Thanks again for your help.
Regards,
default
default

PR #1042, opened 2018-01-25
MolNet update

miaecle:
This PR adds in:

- Updating all models in MoleculeNet benchmark to tensorgraph-based models

- Minor modifications on textcnn model

- New script used automatically generate and update figures for MolNet website


deepchem/data/datasets.py
```diff
@@ -789,11 +786,13 @@ def iterate(dataset, batch_size):
...
 
-        num_local_batches = math.ceil(n_shard_samples / shard_batch_size)
-
         if n_shard_samples == 0:
           cur_shard += 1
+          if batch_size is None:
+            cur_global_batch += 1
           continue
+
+        num_local_batches = math.ceil(n_shard_samples / shard_batch_size)
```
miaecle: I modified here due to some special cases in which n_shard_sampels and shard_batch_size are both 0(corresponding to the situation where a dataset have several shards while one shard has no samples in it) Not sure this will cause any further issues.

deepchem/models/tensorgraph/models/graph_models.py
```diff
@@ -294,7 +305,7 @@ def default_generator(self,
           pad_batches=pad_batches):
 
         feed_dict = dict()
-        if y_b is not None and not predict:
+        if y_b is not None:
```
lilleswing: Removing this breaks predict_on_batch.  I'm trying to figure out a good way to solve it though in my PR here -- https://github.com/deepchem/deepchem/pull/1041

deepchem/data/datasets.py
lilleswing: yikes on creating shards with no samples in it.

miaecle: Yep this is super weird, it happens with scaffold splitting in some small datasets.

deepchem/models/tensorgraph/models/graph_models.py
miaecle: That will be great, I didn't realize this could be destructive.

PR #1046, opened 2018-01-25
[SHOW AND TELL] Keras Model GraphConv and MultiTask

lilleswing:
first stab at what changing TensorGraph Layers to Keras Layers will be like

### DIFF
--- a/examples/tox21/keras_fcnet.py
+++ b/examples/tox21/keras_fcnet.py
@@ -6,9 +6,6 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import numpy as np
-import keras
-from keras import Input, Model
-from keras.layers import Dense, Dropout, Multiply
 import tensorflow as tf
 
 import deepchem as dc
@@ -31,16 +28,16 @@ n_features = 1024
 tox21_tasks, tox21_datasets, transformers = dc.molnet.load_tox21()
 train_dataset, valid_dataset, test_dataset = tox21_datasets
 
-inputs = Input(shape=train_dataset.X[0].shape)
-x = Dense(1000, activation='relu')(inputs)
-x = Dropout(0.25)(x)
+inputs = tf.keras.layers.Input(shape=train_dataset.X[0].shape)
+x = tf.keras.layers.Dense(1000, activation='relu')(inputs)
+x = tf.keras.layers.Dropout(0.25)(x)
 outputs = []
 for i in range(len(tox21_tasks)):
-  predictions = Dense(2, activation='softmax')(x)
+  predictions = tf.keras.layers.Dense(2, activation='softmax')(x)
   outputs.append(predictions)
 
-optimizer = keras.optimizers.Adam(lr=0.001)
-model = Model(inputs=inputs, outputs=outputs)
+optimizer = tf.keras.optimizers.Adam(lr=0.001)
+model = tf.keras.models.Model(inputs=inputs, outputs=outputs)
 model.compile(
     optimizer=optimizer,
     loss='sparse_categorical_crossentropy',
--- a/examples/tox21/keras_fcnet.py
+++ b/examples/tox21/keras_fcnet.py
@@ -1,0 +1,60 @@
+"""
+Script that trains multitask models on Tox21 dataset.
+"""
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+
+import numpy as np
+import keras
+from keras import Input, Model
+from keras.layers import Dense, Dropout, Multiply
+import tensorflow as tf
+
+import deepchem as dc
+
+
+def reshape_y(y):
+  vectors = []
+  for i in range(y.shape[1]):
+    vectors.append(y[:, i])
+  return vectors
+
+
+# Only for debug!
+np.random.seed(123)
+
+metric = dc.metrics.Metric(dc.metrics.roc_auc_score, np.mean)
+
+# Load Tox21 dataset
+n_features = 1024
+tox21_tasks, tox21_datasets, transformers = dc.molnet.load_tox21()
+train_dataset, valid_dataset, test_dataset = tox21_datasets
+
+inputs = Input(shape=train_dataset.X[0].shape)
+x = Dense(1000, activation='relu')(inputs)
+x = Dropout(0.25)(x)
+outputs = []
+for i in range(len(tox21_tasks)):
+  predictions = Dense(2, activation='softmax')(x)
+  outputs.append(predictions)
+
+optimizer = keras.optimizers.Adam(lr=0.001)
+model = Model(inputs=inputs, outputs=outputs)
+model.compile(
+    optimizer=optimizer,
+    loss='sparse_categorical_crossentropy',
+    metrics=['accuracy'])
+model.fit(
+    train_dataset.X,
+    reshape_y(train_dataset.y),
+    batch_size=50,
+    epochs=1,
+    verbose=False,
+    sample_weight=reshape_y(train_dataset.w))
+
+predictions = np.concatenate(model.predict(train_dataset.X), axis=1)
+print(metric.compute_metric(train_dataset.y, predictions))
+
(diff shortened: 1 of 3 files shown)


examples/tox21/keras_fcnet.py
```diff
@@ -0,0 +1,60 @@
...
+from __future__ import unicode_literals
+
+import numpy as np
+import keras
+from keras import Input, Model
+from keras.layers import Dense, Dropout, Multiply
+import tensorflow as tf
+
+import deepchem as dc
+
```
lilleswing: Simple models are really pretty, easy and clean!

examples/tox21/keras_gc.py
```diff
@@ -0,0 +1,253 @@
...
+
+import numpy as np
+import tensorflow as tf
+
+import deepchem as dc
+from deepchem.molnet import load_tox21
+from feat.mol_graphs import ConvMol
+from tensorflow.python.keras import backend as K
+
+sess = tf.Session()
```
lilleswing: more complex models require you to do stuff like K.set_session(session) which sets global state.

examples/tox21/keras_fcnet.py
```diff
@@ -0,0 +1,60 @@
...
+model.compile(
+    optimizer=optimizer,
+    loss='sparse_categorical_crossentropy',
+    metrics=['accuracy'])
+model.fit(
+    train_dataset.X,
+    reshape_y(train_dataset.y),
+    batch_size=50,
+    epochs=1,
+    verbose=False,
```
lilleswing: when verbose is True it gives you all sorts of nice messages about how training is going!

examples/tox21/keras_gc.py
```diff
@@ -0,0 +1,253 @@
...
+  output = tf.nn.softmax(logit)
+  outputs.append(output)
+
+final_loss = tf.reduce_mean(losses)
+train_step = tf.train.AdamOptimizer().minimize(final_loss)
+
+init_op = tf.global_variables_initializer()
+sess.run(init_op)
+
+with sess.as_default():
```
lilleswing: I couldn't use Keras' Model interface because it does strict checking on your inputs and outputs.  For each batch each Input and Output have to have the same length in the zeroth dimension.  

This is impossible for the current featurization scheme of our graph models.  (And I can't think of any easy way to make any given batch have the same zeroth dimension for each input) @peastman @rbharath @miaecle .  We might be able to subclass Model and override the method to not do the strict checking but that could have really bad reprocussions elsewhere in the framework.

lilleswing: Because we can't use the Model class this code gets a lot more ugly really fast.  We still need the tight coupling between the dataset and the Model from TensorGraph.

examples/tox21/keras_gc.py
peastman: I think that's because you're using classes from standalone Keras.

examples/tox21/keras_fcnet.py
```diff
@@ -0,0 +1,60 @@
...
+Script that trains multitask models on Tox21 dataset.
+"""
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+
+import numpy as np
+import keras
+from keras import Input, Model
+from keras.layers import Dense, Dropout, Multiply
```
peastman: These should be tf.keras.

lilleswing: yes it should -- I did this one first before the GC one

examples/tox21/keras_gc.py
lilleswing: keras_gc only uses classes from the Tensorflow implementation.  

I had to do K.set_session() and K.set_learning_phase() in order for tf.keras.layers.BatchNormalization to work, or I could of done what TensorGraph does and create a tf.variable for training.

```python
# implementation of BatchNormalization#call
  def call(self, inputs, training=None):
    if training is None:
      training = K.learning_phase()
    output = super(BatchNormalization, self).call(inputs, training=training)
    if training is K.learning_phase():
      output._uses_learning_phase = True  # pylint: disable=protected-access
    return output
```

lilleswing: yeah you are right -- we can probably get around this in all cases.

examples/tox21/keras_gc.py
peastman: Are you sure about that?  I'm looking through the code for `Model.fit_generator()` and I don't see any restriction like that.  On the contrary, for every batch it separately looks up the batch size and logs it: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/_impl/keras/engine/training.py#L2160-L2169.  Can you post a simple example that shows the problem?

lilleswing: I was looking at fit not fit_generator().  Give me some time to rebundle the keras_gc more in keras land (replace tf.placeholder with tf.keras.layers.Input, etc) and see if we can use fit_generator().

peastman: `fit()` looks like a more limited function.  You give it numpy arrays containing the entire dataset, and it breaks them up into batches for you.  `fit_generator()` lets you decide how to break them into batches.

lilleswing: it is still happening.  Check out keras_gc_full_keras.py
```python
Epoch 1/1
  File "/home/leswing/Documents/deepchem/examples/tox21/keras_gc_full_keras.py", line 233, in <module>
    model.fit_generator(generator(train_dataset), 10)
  File "/home/leswing/miniconda3/envs/deepchem/lib/python3.5/site-packages/tensorflow/python/keras/_impl/keras/engine/training.py", line 2076, in fit_generator
    x, y, sample_weight=sample_weight, class_weight=class_weight)
  File "/home/leswing/miniconda3/envs/deepchem/lib/python3.5/site-packages/tensorflow/python/keras/_impl/keras/engine/training.py", line 1781, in train_on_batch
    check_batch_axis=True)
  File "/home/leswing/miniconda3/envs/deepchem/lib/python3.5/site-packages/tensorflow/python/keras/_impl/keras/engine/training.py", line 1410, in _standardize_user_data
    _check_array_lengths(x, y, sample_weights)
  File "/home/leswing/miniconda3/envs/deepchem/lib/python3.5/site-packages/tensorflow/python/keras/_impl/keras/engine/training.py", line 229, in _check_array_lengths
    [x.shape for x in inputs]))
ValueError: All input arrays (x) should have the same number of samples. Got array shapes: [(1279, 75), (11, 2), (1279, 1), (313, 1), (583, 2), (341, 3), (41, 4), (0, 5), (0, 6), (0, 7), (0, 8), (0, 9), (0, 10)]
```

lilleswing:
We would have to remove this line
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/_impl/keras/engine/training.py#L1447
to get around the input checks.

I don't see a way to do it with flag passing.

lilleswing:
forking and installing from our own conda-forge channel is hard because we want to use the version shipped with tensorflow also :(

lilleswing:
p.s. it passes simple smoke test of training and predicting when I remove the input validation checks.

peastman:
Maybe I misunderstood exactly what issue you're describing.  You're saying that within a single batch, you want different input arrays to have different numbers of samples?  What is the situation where that comes up?

lilleswing:
GraphConvs inputs are per atom features and a description of multiple
graphs.  They are either sample per atom or sample per atom with degree x.

The y values are properties to be predicted on molecules.  That is value
per molecule.

I guess we could expand_dims(axis=0) in the featurization, and then squeeze
it out in the first layer to get around this issue.

peastman:
I think all of that would be considered a single sample?  It's a batch size of 1, with a whole lot of data for each sample.

examples/tox21/keras_fcnet.py
rbharath: Yes, agreed! It's also great since anyone who knows TensorFlow can immediately see how to train DeepChem models.

PR #1057, opened 2018-01-29
Website Updates

lilleswing:
Adds
* About Page
* Top Level Link to Blog
* Top Level Link to Tutorials
* Docs goes directly to our pydocs instead to a pseudo about page.

### DIFF
--- a/website/about.html
+++ b/website/about.html
@@ -17,7 +17,7 @@
 
 <header>
 
-    <div class="logo"><a href="#" class="pull-left"><img src="./static/img/deepchem_logo.png" height="100"></a></a></div>
+    <div class="logo"><a href="./index.html" class="pull-left"><img src="./static/img/deepchem_logo.png" height="100"></a></a></div>
 
     <ul class="primary-nav">
       <li><a  href="./about.html">About</a></li>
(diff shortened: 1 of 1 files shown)


rbharath:
@lilleswing Do you have a rendered version of the updated website we can take a look at?

rbharath:
Just saw that the website looks to have been updated already. I think there are a couple CSS problems though. https://deepchem.io/docs/notebooks/index.html is rendering without CSS for me.

lilleswing:
Not 100% sure what is happening -- it was fine yesterday.  It is also fine on the direct site.
https://s3-us-west-1.amazonaws.com/deepchem.io/docs/notebooks/index.html

Which makes me think it has to do with the redirect again.

lilleswing:
So I believe that something changed with our GoDaddy registration @rbharath

lilleswing:
The MIME type of "text/css" is being lost on _static/bootstrap_sphinx.css accross the proxy we have. This is causing bootstrap_sphinx.css to be read as a text file and not loading in the @import on its first line.

rbharath:
You want to recursively set mime type. For example see solution in

https://github.com/deepchem/deepchem/issues/777

The file linked to in there was deleted at some point (which it shouldn't have been). Might need to look into history to find it.

lilleswing:
We are still doing that.  It is set correctly on amazon -- hence it working correctly at the direct location.
https://github.com/deepchem/deepchem/blob/master/devtools/jenkins/push-docs-to-s3.py

selection_057


The MIME type is not making it through the proxy.

lilleswing:
That was a really tricky problem!  I'm going to leave some notes in case I ever need to do something like this again for the website.

The website is a S3 bucket, uses route 53 for DNS and cloudfront for CDN.
The DNS points to cloudfront which then loads from the S3 bucket directly so we have faster page loads across the country.

The MIME type was reset in S3 and propagated to the direct website.  However we had to bust the cache in CDN so that it could recursively grab the css in _static/bootstrp_sphinx.css.

selection_058

Why the recursive mimetype magic didn't work in the first place is unknown.
Why we even updated the file in S3 when it should only update on diffs is unknown.

tl;dr
The site has css again.

rbharath:
@lilleswing Nice debugging!

LGTM

PR #1099, opened 2018-02-18
Fixed IndiceSplitter

abster12:
Assigned a copy of the list self.valid_indices to valid_test.

### DIFF
--- a/deepchem/splits/splitters.py
+++ b/deepchem/splits/splitters.py
@@ -753,7 +753,7 @@ class IndiceSplitter(Splitter):
       self.valid_indices = []
     if self.test_indices is None:
       self.test_indices = []
-    valid_test = self.valid_indices
+    valid_test = list(self.valid_indices)
     valid_test.extend(self.test_indices)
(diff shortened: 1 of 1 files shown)


deepchem/splits/splitters.py
```diff
@@ -753,7 +753,7 @@ def split(self,
       self.valid_indices = []
     if self.test_indices is None:
       self.test_indices = []
-    valid_test = self.valid_indices
+    valid_test = list(self.valid_indices)
```
lilleswing: can we get a unit test?

abster12: A unit test asserting that both the values are same?

lilleswing: so the way I am reading this is

There used to be a mutation in the self.valid_indices which made it also contain the self.test_indices.  We want to have a unit test showing that would fail without this fix and pass with this fix.

That is a unit test which shows that the valid_indices no longer has the self.test_indices in it.

PR #1105, opened 2018-02-21
8x faster graph normalization

ueser:
2 changes:
1. numpy matrix inversion is 2x faster than scipy inversion
2. For diagonal matrices, np.sqrt and scipy.linalg.sqrtm give the same result because there are no interaction terms. But np.sqrt is 250x faster.


lilleswing:
LGTM thanks for the contribution!

PR #1141, opened 2018-03-07
[WIP]Apply split==None to all the load functions

nitinprakash96:


examples/notebooks/deepchem_tensorflow_eager.ipynb
```diff
@@ -0,0 +1,32 @@
+{
```
lilleswing: please remove this notebook from this PR.

nitinprakash96:
@lilleswing how do we add this feature in qm7, qm8 and qm9 dataset load modules? Because we are not really applying transformers before splitters.

lilleswing:
@nitinprakash96 The best way would be to raise a ValueError() when None is specified.

nitinprakash96:
@lilleswing let me know if the changes are okay

deepchem/molnet/load_function/bbbp_datasets.py
```diff
@@ -47,19 +50,22 @@ def load_bbbp(featurizer='ECFP', split='random', reload=True):
...
-  splitters = {
-      'index': deepchem.splits.IndexSplitter(),
-      'random': deepchem.splits.RandomSplitter(),
-      'scaffold': deepchem.splits.ScaffoldSplitter()
-  }
-  splitter = splitters[split]
-  train, valid, test = splitter.train_valid_test_split(dataset)
+  if split == None:
+    return bbbp_tasks, (dataset, None, None), transformers
+  else:
```
lilleswing: this else is unnecessary and adds an extra layer of indentation.  Please remove.

lilleswing: Here and elsewhere.  As a general rule of thumb extra layers of indentation are bad.  It makes the code harder to read and makes the programmer store state in the if/else tree in their mind.

deepchem/molnet/load_function/chembl_datasets.py
```diff
@@ -103,18 +106,21 @@ def load_chembl(shard_size=2000,
...
+        'index': deepchem.splits.IndexSplitter(),
+        'random': deepchem.splits.RandomSplitter(),
+        'scaffold': deepchem.splits.ScaffoldSplitter()
+    }
 
-  if split in splitters:
-    splitter = splitters[split]
-    print("Performing new split.")
-    train, valid, test = splitter.train_valid_test_split(dataset)
+    if split in splitters:
```
lilleswing: why are we checking if split is in splitters here?  We have already done the None check.  If it is not in splitters we will raise a really weird exception about not initializing train, valid, test.  Where as if you don't do the check the exception raised will make far more sense to the programmer.

lilleswing: I understand you didn't put in this code but we should leave it better than we found it now that we are here :smile:

deepchem/molnet/load_function/qm9_datasets.py
```diff
@@ -69,25 +72,28 @@ def load_qm9(featurizer='CoulombMatrix', split='random', reload=True):
...
-      deepchem.trans.NormalizationTransformer(
-          transform_y=True, dataset=train_dataset)
-  ]
-  for transformer in transformers:
-    train_dataset = transformer.transform(train_dataset)
-    valid_dataset = transformer.transform(valid_dataset)
-    test_dataset = transformer.transform(test_dataset)
+  if split == None:
+    raise ValueError()
+  else:
```
lilleswing: This else falls into the same category as before.  There is no need for it as the raise from the if already exits the method.

lilleswing:
Looks great!  Just a couple of code nice-ities left and this will be good to go!

lilleswing:
LGTM are you waiting on anything else @nitinprakash96 ?

nitinprakash96:
@lilleswing I think we are good at this point.

PR #1147, opened 2018-03-13
Enable Atom-Based Properties Stored in SDF Files to be used in the ConvMol Featurizer

KenBorrelli:
Enable Atom-Based Properties Stored in SDF Files to be used in the ConvMol Featurizer
Addresses issue 1146

### DIFF
--- a/deepchem/feat/graph_features.py
+++ b/deepchem/feat/graph_features.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
 
@@ -264,7 +265,10 @@ def find_distance(a1, num_atoms, canon_adj_list, max_distance=7):
 class ConvMolFeaturizer(Featurizer):
   name = ['conv_mol']
 
-  def __init__(self, master_atom=False, use_chirality=False):
+  def __init__(self,
+               master_atom=False,
+               use_chirality=False,
+               atom_properties=None):
     """
     Parameters
     ----------
@@ -274,7 +278,20 @@ class ConvMolFeaturizer(Featurizer):
       the molecule.  This technique is briefly discussed in
       Neural Message Passing for Quantum Chemistry
       https://arxiv.org/pdf/1704.01212.pdf
-
+    use_chirality: Boolean
+      if true then make the resulting atom features aware of the 
+      chirality of the molecules in question
+    atom_properties: list of string or None
+      properties in the RDKit Mol object to use as additional
+      atom-level features in the larger molecular feature.  If None,
+      then no atom-level properties are used.  Properties should be in the 
+      RDKit mol object should be in the form 
+      atom XXXXXXXX NAME
+      where XXXXXXXX is a zero-padded 8 digit number coresponding to the
+      zero-indexed atom index of each atom and NAME is the name of the property
+      provided in atom_properties.  So "atom 00000000 sasa" would be the
+      name of the molecule level property in mol where the solvent 
+      accessible surface area of atom 0 would be stored. 
 
     Since ConvMol is an object and not a numpy array, need to set dtype to
     object.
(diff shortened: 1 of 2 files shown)


deepchem/feat/graph_features.py
```diff
@@ -274,20 +277,60 @@ def __init__(self, master_atom=False, use_chirality=False):
...
 
     Since ConvMol is an object and not a numpy array, need to set dtype to
     object.
     """
     self.dtype = object
     self.master_atom = master_atom
     self.use_chirality = use_chirality
+    self.atom_properties = atom_properties
+
+  def get_atom_properties(self, atom):
```
lilleswing: _get_atom_properties -- no reason to expose to outside world.

deepchem/feat/graph_features.py
```diff
@@ -274,20 +277,60 @@ def __init__(self, master_atom=False, use_chirality=False):
...
+    requested when initializing the featurize.  See the __init__ of the
+    class for a full description of the names of the properties
+    
+    Parameters
+    ----------
+    atom: RDKit.rdchem.Atom
+      Atom to get the properties of
+    returns a numpy lists of floats of the same size as self.atom_properties
+    """
+    values = []
```
lilleswing: short circuit for none case of self.atom_properties.  This is also the reason you are failing all of the tests.

deepchem/feat/graph_features.py
```diff
@@ -274,20 +277,60 @@ def __init__(self, master_atom=False, use_chirality=False):
...
+      name of the molecule level property in mol where the solvent 
+      accessible surface area of atom 0 would be stored. 
 
     Since ConvMol is an object and not a numpy array, need to set dtype to
     object.
     """
     self.dtype = object
     self.master_atom = master_atom
     self.use_chirality = use_chirality
+    self.atom_properties = atom_properties
```
lilleswing: defaulting to empty list would fix the tests as is

lilleswing: Also can you wrap the sequence in a list() I feel like users will pass in an iterable and then get upset when it iterates in non deterministic order.

deepchem/models/tensorgraph/models/graph_models.py
```diff
@@ -694,14 +699,15 @@ def __init__(self,
...
+    self.number_atom_features = number_atom_features
     super(GraphConvTensorGraph, self).__init__(**kwargs)
     self.build_graph()
 
   def build_graph(self):
     """
-            Building graph structures:
-            """
-    self.atom_features = Feature(shape=(None, 75))
+    Building graph structures:
```
lilleswing: Can you add a test for this in test_graph_models?  Something like
```python
  def test_graph_conv_atom_features(self):
    tasks, dataset, transformers, metric = self.get_dataset(
        'regression', 'Raw', num_tasks=1)

    atom_feature_name = 'feature'
    y = []
    for mol in dataset.X:
      atom_features = []
      for atom in mol.GetAtoms():
        val = np.random.normal()
        mol.SetProp("atom %08d %s" % (atom.GetIdx(), atom_feature_name), str(val))
        atom_features.append(np.random.normal())
      y.append(np.sum(atom_features))

    featurizer = ConvMolFeaturizer(atom_properties=[atom_feature_name])
    X = featurizer.featurize(dataset.X)
    dataset = deepchem.data.NumpyDataset(X, np.array(y))
    batch_size = 50
    model = GraphConvTensorGraph(
        len(tasks),
        additional_atom_features=1,
        batch_size=batch_size,
        mode='regression')

    model.fit(dataset, nb_epoch=1)
    y_pred1 = model.predict(dataset)
    model.save()

    model2 = TensorGraph.load_from_dir(model.model_dir)
    y_pred2 = model2.predict(dataset)
    self.assertTrue(np.all(y_pred1 == y_pred2))
```

deepchem/feat/graph_features.py
lilleswing: defaulting to empty list is better than defaulting to None -- then you won't need the short circuit

deepchem/feat/graph_features.py
```diff
@@ -274,20 +277,60 @@ def __init__(self, master_atom=False, use_chirality=False):
...
+    returns a numpy lists of floats of the same size as self.atom_properties
+    """
+    values = []
+    for prop in self.atom_properties:
+      mol_prop_name = "atom %08d %s" % (atom.GetIdx(), prop)
+      try:
+        values.append(float(atom.GetOwningMol().GetProp(mol_prop_name)))
+      except KeyError:
+        raise KeyError("No property %s found in %s in %s" %
+                       (mol_prop_name, atom.GetOwningMol(), self))
```
lilleswing: This is going to hive you a memory position for the rdkit mol but that is fine.

lilleswing:
I'm going to close this as I'm driving the PR from #1156 now.

PR #1149, opened 2018-03-13
[WIP] Scscore model

abster12:
This is just the model architecture I was trying. I havent done anything much to it due to my exams. I will try to finish it as soon as possible

### DIFF
--- a/deepchem/models/tensorgraph/models/scscore.py
+++ b/deepchem/models/tensorgraph/models/scscore.py
@@ -1,23 +1,28 @@
 import deepchem as dc
 import tensorflow as tensorflow
-from deepchem.models.tensorgraph.layers import Layer,Input,Dense,Hingeloss,Label,Feature
-from deepchem.models.tensorgraph.layers import ReduceMean,InputFifoQueue,ReLU
+from deepchem.models.tensorgraph.layers import Layer, Input, Dense, Hingeloss, Label, Feature
+from deepchem.models.tensorgraph.layers import ReduceMean, InputFifoQueue, ReLU
+
 
 class Scscore(TensorGraph):
-	tg=dc.models.TensorGraph(tensorboard=True, model_dir='/tmp/scscore',use_queue=True)
+  tg = dc.models.TensorGraph(
+      tensorboard=True, model_dir='/tmp/scscore', use_queue=True)
 
+  Input1 = InputFifoQueue()
 
-	Input1=InputFifoQueue()
+  hidden_1 = Dense(out_channels=300, in_layer=Input1, activation_fn=tf.nn.relu)
+  hidden_2 = Dense(
+      out_channels=300, in_layer=hidden_1, activation_fn=tf.nn.relu)
+  hidden_3 = Dense(
+      out_channels=300, in_layer=hidden_2, activation_fn=tf.nn.relu)
+  hidden_4 = Dense(
+      out_channels=300, in_layer=hidden_3, activation_fn=tf.nn.relu)
+  hidden_5 = Dense(
+      out_channels=300, in_layer=hidden_4, activation_fn=tf.nn.relu)
 
-	hidden_1=Dense(out_channels=300,in_layer=Input1,activation_fn=tf.nn.relu)
-	hidden_2=Dense(out_channels=300,in_layer=hidden_1,activation_fn=tf.nn.relu)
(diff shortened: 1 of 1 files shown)


deepchem/models/tensorgraph/models/scscore.py
```diff
@@ -0,0 +1,28 @@
...
+import tensorflow as tensorflow
+from deepchem.models.tensorgraph.layers import Layer, Input, Dense, Hingeloss, Label, Feature
+from deepchem.models.tensorgraph.layers import ReduceMean, InputFifoQueue, ReLU
+
+
+class Scscore(TensorGraph):
+  tg = dc.models.TensorGraph(
+      tensorboard=True, model_dir='/tmp/scscore', use_queue=True)
+
+  Input1 = InputFifoQueue()
```
lilleswing: You don't specify InputFifoQueue, instead use Feature

deepchem/models/tensorgraph/models/scscore.py
```diff
@@ -0,0 +1,28 @@
...
+  hidden_3 = Dense(
+      out_channels=300, in_layer=hidden_2, activation_fn=tf.nn.relu)
+  hidden_4 = Dense(
+      out_channels=300, in_layer=hidden_3, activation_fn=tf.nn.relu)
+  hidden_5 = Dense(
+      out_channels=300, in_layer=hidden_4, activation_fn=tf.nn.relu)
+
+  label = Label(shape=(300,))
+
+  loss = Hingeloss(in_layers=[label, hidden_5])
```
lilleswing: How are you going to send pairs of molecules through this?  The whole architecture is based on pairwise molecules where one is "more complex" than the other.

abster12:
I have used the shared layers but I was confused about where to add the reactant and product layers. Is it supposed to be added at the end of the dense layers?

deepchem/models/tensorgraph/models/scscore.py
```diff
@@ -0,0 +1,39 @@
...
+      out_channels=300, in_layers=[dense_reactant_2], activation_fn=tf.nn.relu)
+  dense_product_3 = dense_reactant_3.shared(in_layers=[dense_product_2])
+  dense_reactant_4 = Dense(
+      out_channels=300, in_layers=[dense_reactant_3], activation_fn=tf.nn.relu)
+  dense_product_4 = dense_reactant_4.shared(in_layers=[dense_product_3])
+  dense_reactant_5 = Dense(
+      out_channels=300, in_layers=[dense_reactant_4], activation_fn=tf.nn.relu)
+  dense_product_5 = dense_reactant_5.shared(in_layers=[dense_product_4])
+
+  output_dense = Add(in_layers=[dense_reactant_5, dense_product_5])
```
lilleswing: I don't think this is what you want to do.

abster12:
In the scscore paper the author says that they use 2 models to calculate the 2 scores seperately for reactant and product and then the difference between their scores are evaluated. 
So is my mistake that I dont have to add the layers and continue with the 2 layers for reactant and product and the evaluate the score for them and then calculate the difference from the scores?

deepchem/models/tensorgraph/models/scscore.py
abster12: In the scscore paper the author says that they use 2 models to calculate the 2 scores seperately for reactant and product and then the difference between their scores are evaluated.
So is my mistake that I dont have to add the layers and continue with the 2 layers for reactant and product and the evaluate the score for them and then calculate the difference from the scores?

deepchem/models/tensorgraph/models/scscore.py
```diff
@@ -0,0 +1,40 @@
...
+  dense_reactant_5 = Dense(
+      out_channels=300, in_layers=[dense_reactant_4], activation_fn=tf.nn.relu)
+  dense_product_5 = dense_reactant_5.shared(in_layers=[dense_product_4])
+
+  output_reactant = Sigmoid(in_layers=[dense_reactant_5])
+  output_product = Sigmoid(in_layers=[dense_product_5])
+  output = tf.subtract(output_product, output_reactant)
+  tg.add_output(output)
+
+  label = Label(shape=(None, 1))
```
abster12: I wanted to ask of we needed the label in this one because the paper discusses how they avoid the ground truth and instead learn from ranked pairs.

PR #1163, opened 2018-03-17
Fix syntax error in graph_features.py

n3011:
fix syntax error


lilleswing:
Nice catch!

To finish the review please run our code-formatter.

```bash
pip install yapf==0.20.0
yapf -i deepchem/feat/graph_features.py
```

n3011:
closing this one; as the fix is already included in my second PR

Issue #1188, opened 2018-03-27
MPNN Edge Network Implementation


WesleyyC:
For the Edge Network in MPNN, the paper is doing A x h, but the implementation is doing elementwise product with broadcasting and has to do a reduce sum to get the right size:

https://github.com/deepchem/deepchem/blob/281a1e050ac5ef7ce4c8d481a96c44e8816e88c6/deepchem/models/tensorgraph/graph_layers.py#L928

I think a batch mode matmul makes more sense. Therefore, to replace this line:

```python
out = tf.squeeze(tf.matmul(self.A, out), axis=2)
```

I have the fix ready and would be happy to create a PR if this makes sense to you guys.

miaecle:
@WesleyyC Hi, thanks for raising the issue! Since `tf.matmul` supports this kind of operation, I totally agree with this change. Please feel free the initialize a PR!

WesleyyC:
@miaecle please see the PR. Thanks!

WesleyyC:
Fix is merged

PR #1195, opened 2018-03-28
Use tf.matmul for batch matrix multiplication in MPNN's Edge Network

WesleyyC:
Please see #1188 for details. cc @miaecle


lilleswing:
LGTM thanks for the enhancement!

PR #1196, opened 2018-03-29
Move Tensor Graph old model deprecation warning to constructor

mlgill:
Per [this](https://github.com/deepchem/deepchem/pull/1181#issuecomment-376938610) comment, I've moved the Tensor Graph old-style model naming to the constructor for the class so that error warnings aren't generated upon importing deepchem.


lilleswing:
LGTM thanks for the quick fix!

PR #1227, opened 2018-04-18
Fix for graph conv regression

miaecle:
This PR fixes

- Missing `transformers` argument in `evaluate` function (causing wrong metric values for RMSE, MAE, etc.)

- Wrong dataset in `test_graph_conv_model`


peastman:
> Wrong dataset in test_graph_conv_model

Wrong in what sense?  It just uses the dataset to get a list of molecules, then generates fake y and w arrays for it.

miaecle:
@peastman The get_dataset function loads delaney for classification test, together with the Normalization transformer. I think the transformer causes an error in the evaluation after I make the change so I simply switch it to another classification dataset.

PR #1229, opened 2018-04-19
Dropout for DAG

miaecle:
This PR adds:

- Fix some doc strings and arguments in `DAGLayer`

- Initial support for dropout in DAG model

Due to the special structure of DAG, the dropout structure added here is a stepwise process that happens on each step(propagate through one directed edge). Another approach is only doing dropout on  the outputs, or equivalently the last step of propagation. Comments and suggestions on the potential structure choice here are very welcomed!

### DIFF
--- a/deepchem/models/tensorgraph/graph_layers.py
+++ b/deepchem/models/tensorgraph/graph_layers.py
@@ -581,7 +581,7 @@ class DAGLayer(Layer):
           Number of features listed per atom.
         max_atoms: int, optional
           Maximum number of atoms in molecules.
-        layer_sizes: list of int, optional(default=[1000])
+        layer_sizes: list of int, optional(default=[100])
           Structure of hidden layer(s)
         init: str, optional
           Weight initialization for filters.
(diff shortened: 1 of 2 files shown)


deepchem/models/tensorgraph/graph_layers.py
```diff
@@ -698,11 +699,13 @@ def create_tensor(self, in_layers=None, set_tensors=True, **kwargs):
       self.out_tensor = out_tensor
     return out_tensor
 
-  def DAGgraph_step(self, batch_inputs, W_list, b_list):
+  def DAGgraph_step(self, batch_inputs, W_list, b_list, **kwargs):
     outputs = batch_inputs
     for idw, W in enumerate(W_list):
       outputs = tf.nn.xw_plus_b(outputs, W, b_list[idw])
       outputs = self.activation(outputs)
+      if 'training' in kwargs and kwargs['training'] == 1.0 and not self.dropout is None:
```
peastman: `training` will often be a tensor, not a number, so this comparison will fail.  Just multiply the dropout rate by it and don't worry about the value (like in the Dropout layer).

miaecle: Good point, just made the change!

deepchem/models/tensorgraph/models/graph_models.py
```diff
@@ -381,6 +384,12 @@ def __init__(self,
               Number of features for atom in the graph
             n_outputs: int, optional
               Number of features for each molecule
+            layer_sizes: list of int, optional
+              Structure of hidden layer(s) in DAG propagation step
```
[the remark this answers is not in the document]
miaecle: Sure, added a more detailed explanation here.

peastman:
Looks good.

PR #1259, opened 2018-05-14
Removed predict_proba()

### LINKED ISSUE #1214: What is predict_proba() for?

peastman:
Fixes #1214.

Please look this over carefully.  It was challenging to get this working, and I ran into some inconsistent code that had to be fixed.  For example, the top level Model class had the hardcoded assumption that model outputs would always be exactly two dimensional, of shape (samples, tasks).  Some of the routines in multitask.py had the same assumption.

### DIFF
--- a/deepchem/models/models.py
+++ b/deepchem/models/models.py
@@ -80,18 +80,6 @@ class Model(BaseEstimator):
     raise NotImplementedError(
         "Each model is responsible for its own predict_on_batch method.")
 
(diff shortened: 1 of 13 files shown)


lilleswing:
This is much cleaner.  LGTM

PR #1267, opened 2018-05-23
Hashable Featurizers

lilleswing:
Implement the methods needed to make ConvMolFeaturizer and CircularFingerPrint hashable.

This enables functions which take these featurizers as args to use the functools lru_cache.


lilleswing:
Imagine running Gaussian Processes over hyper parameters for a supervised learning problem, with featurization as a hyper parameter.

Eventually the featurization will become more stable and we won't need to keep incurring the cost of featurization.

PR #1272, opened 2018-05-24
[WIP] U-Net implementation (Phase 1)

pvskand:
Initial implementation of U-Net architecture.
This consists of the model of U-Net.
This PR aims to complete the following:

- [x] U-Net architecture
- [x] Metrics for U-Net architecture
Tutorial to use U-Net architecture in `examples/notebooks` will be covered in the next PR after implementation of `DataTransforms` API.

### DIFF
--- a/deepchem/models/tensorgraph/models/unet.py
+++ b/deepchem/models/tensorgraph/models/unet.py
@@ -13,58 +13,65 @@ from deepchem.models.tensorgraph.layers import Conv2D, MaxPool2D, Conv2DTranspos
 from deepchem.models import TensorGraph
 
 class UNet(TensorGraph):
-
- """
- U-Net architecture implementation.
- Parameters
- ----------
- img_rows : int
-	 number of rows of the image.
- img_cols: int
-	 number of columns of the image.
- """
-
-	def __init__(self,img_rows=512, img_cols=512, **kwargs):
+# """
+# U-Net architecture implementation.
+# Parameters
+# ----------
+# img_rows : int
+#  number of rows of the image.
+# img_cols: int
+#  number of columns of the image
+# """
+    def __init__(self,img_rows=512, img_cols=512, **kwargs):
 		super(UNet, self).__init__(use_queue=False, **kwargs)
 		self.img_cols = img_cols
 		self.img_rows = img_rows
 
-	def unet_model(self):
-		model = dc.models.TensorGraph()
+    def unet_model(self):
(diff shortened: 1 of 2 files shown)


peastman:
Thanks for the contribution!

I'm a little confused about how this is meant to be used, since you've put the actual model building into a separate `unet_model()` method.  Is the user supposed to call that?  Why not just build the model in the constructor?

You've hardcoded a lot of properties of the model: the number of layers, the number of output channels for each layer, etc.  It would be good to make those user configurable, since they'll be problem specific.  For example, larger input images will generally require more layers than smaller images will.

pvskand:
@peastman thanks for the reviews.
1) I can put the model in the `__init__` itself and remove the `unet_model`.
2) My proposal consisted of U-Net architecture and this has a fixed architecture with the configuration that I have coded. Since there is no use of `Dense Layers`, I have not hardcoded the `img_cols` and `img_rows`. I would also like to hear from @lilleswing and @rbharath on hardcoding the layers inputs/outputs.
@peastman  what do you suggest? How should I keep it configurable? Like there are multiple layers should all the layer's output be configurable by the user?

peastman:
A simple way of doing it would be to have a constructor argument `layer_channels=[64, 128, 256, 512, 1024]`.  The default value would match what's in the paper, but the user could provide a different value to change both the number of layers and the number of output channels in each layer.

I compared your code to what's shown in figure 1 of the paper, and I notice a couple of difference.  First, you have half as many output channels at each layer as they do.  Second, they perform two conv2d operations between each set of pooling or upscaling operations, but you only do one.  Were those differences intentional?

deepchem/models/tensorgraph/models/unet.py
```diff
@@ -0,0 +1,74 @@
...
+        img_cols: int
+         number of columns of the image
+    """
+    def __init__(self,img_rows=512, img_cols=512, model=dc.models.TensorGraph(), **kwargs):
+        super(UNet, self).__init__(use_queue=False, **kwargs)
+        self.img_cols = img_cols
+        self.img_rows = img_rows
+        self.model = dc.models.TensorGraph()
+
+        input = Feature(shape=(None, self.img_rows, self.img_cols))
```
lilleswing: input is a python keyword.  Can we also put the graph building in a method
```python
def build_graph(self):
```
like the Graph Models?

This will help for consistency.

Issue #1281, opened 2018-05-31
Graph conv Weights_28/Placeholder dimension mismatch


timaro:
**Description:**

Trying to train a GraphConvTensorGraph, I get the following exception with DeepChem HEAD:

```
File "/lib/python3.6/site-packages/deepchem/models/tensorgraph/tensor_graph.py", line 153, in fit
    max_checkpoints_to_keep, checkpoint_interval, restore, submodel)
  File "/lib/python3.6/site-packages/deepchem/models/tensorgraph/tensor_graph.py", line 257, in fit_generator
    fetched_values = self.session.run(fetches, feed_dict=feed_dict)
  File "/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 905, in run
    run_metadata_ptr)
  File "/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 1113, in _run
    str(subfeed_t.get_shape())))
ValueError: Cannot feed value of shape (100,) for Tensor 'Weights_28/Placeholder:0', which has shape '(?, 1)'
```

This exception does not happen on the same code, when using the most current release version.

- DeepChem Version: HEAD (2.0.1.dev250; commit 1b171da623799b47871b95349b96ff2b139e3f41)
- Platform: Ubuntu 16.04
- Relevant libs: tensorflow-gpu  1.6.0

peastman:
Can you provide example code to reproduce the problem?  TensorGraph is supposed to automatically add or remove dimensions of size 1 as necessary to make the values in the dataset match the expected inputs.

timaro:
I'm doing nothing special: creating a GraphConvTensorGraph and calling
fit():

model = GraphConvTensorGraph(1, batch_size=100, mode="classification")
model.fit(data, nb_epoch=10)

The data instance is a NumpyDataset with dimensions:

data.X.shape -> (9910,)
data.y.shape -> (9910, 1)

Again, the training works correctly on release 2.0. It's only HEAD that is
failing. I see that you made a commit on the NumpyDataset file in the past
two days related to reshaping, so perhaps that's related...

timaro:
Yes, reverting commit 2fba2bbcd68e7a145e7fdd584f8b19b75d875b95 fixes the problem.

peastman:
[...]

PR #1283, opened 2018-05-31
restore numpy array reshaping for Y & w in NumpyDataset init

timaro:
Ensure that NumpyDataset Y & w values are reshaped to tensor rank 2.

Resolves issue #1281

### DIFF
--- a/deepchem/data/datasets.py
+++ b/deepchem/data/datasets.py
@@ -323,6 +323,13 @@ class NumpyDataset(Dataset):
       y = np.array(y)
     if not isinstance(w, np.ndarray):
       w = np.array(w)
+
+    if len(y.shape) == 1:
+      y = np.reshape(y, (n_samples, -1))
+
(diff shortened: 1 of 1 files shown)


peastman:
This is not a correct thing to do.  We specifically removed it because it's not correct.  Datasets can have arrays with any dimensionality, including 1.

timaro:
A numpy array of shape (N, -1) is equivalent to shape (N,). What this change does is restore the rank-two tensor shape when a rank-one tensor is passed in, which is a perfectly safe thing to do.

To make the code fully general, I can change the commit to use len(Y.shape) < 2, which solves the problem and allows inputs of any dimension.

timaro:
I should be clearer: change the *condition* on the if statements to use len(Y.shape) < 2 and len(w.shape) < 2. This corrects the shape mis-match issue in the trivial case of 1D vectors without loss of generality.

PR #1311, opened 2018-07-13
OntologyModel

peastman:
This is the beginning of implementing #1245.  It's a hierarchical model whose structure is determined by some known organization, such as the categorization of genes in Gene Ontology.  At the moment you have to define the hierarchy yourself, but I'll look into automatically building it for GO.


peastman:
Note that I added a `default_outputs` field to TensorGraph.  The loss function for this model encourages every node in the hierarchy to independently try to predict the labels, and the user may want to query the predictions for internal nodes.  That's part of what makes the model interpretable.  So I had to add all of them to the list of outputs so `create_tensor()` would get called on them.  But the default for `predict()` should still be to return only the top level prediction, not a list of predictions from every node in the hierarchy.  That's why I added `default_outputs`.

peastman:
Let's go ahead and merge this.  Eventually it would be good to have a full example of using the model to analyze genomic data.  That's going to be a big project to create, though, and since no one is currently asking to use this model, I don't even know what kind of data would be best.

PR #1340, opened 2018-08-02
[WIP]Test graph layers

abster12:
I started with writing the tests for the graph layers. What I wanted to ask was that is this the way the tests are supposed to be or some changes have to be made?


deepchem/models/tensorgraph/tests/test_graph_layers.py
```diff
@@ -0,0 +1,23 @@
...
+from deepchem.data import NumpyDataset, pad_features
+from deepchem.feat.mol_graphs import ConvMol
+from deepchem.metrics import to_one_hot
+from deepchem.models.tensorgraph.graph_layers import WeaveGather, DTNNEmbedding, DTNNStep, DTNNGather, DAGLayer, DAGGather, DTNNExtract, MessagePassing, SetGather, WeaveLayer, EdgeNetwork, GatedRecurrentUnit
+
+from tensorflow.python.framework import test_util
+
+
+class TestGraphLayers(test_util.TensorFlowTestCase):
+    def Weavegather(self):
```
peastman: Rename the method to `test_weave_gather`.  nosetest only treats a method as a test if its name begins with "test".

peastman:
That looks like the start of a test case.  Now you need to check whether the result it produced is correct.

peastman:
You can just invoke the layer directly, passing in whatever inputs you need.  Here's a good example of a test case for a layer that passes in known inputs and checks that the output is correct: https://github.com/deepchem/deepchem/blob/master/deepchem/models/tensorgraph/tests/test_layers.py#L908-L921

abster12:
I was not able to figure out how to get the expected outputs of the layers, also are the tests going to look like this or the way I have used the layers is wrong.
Also I was looking at the other layers they have so many layers as the input how do I write a test for those?

PR #1370, opened 2018-08-13
Replacing MaxPool with AvgPool in Resnet50

pvskand:
Previously since DeepChem did not have Average pooling, I had used Max pool instead. Now replacing it with `AvgPool2D`.


pvskand:
Looks like a stochastic build fail. Correct me if I am wrong : https://travis-ci.org/deepchem/deepchem/jobs/415439355#L3231

lilleswing:
LGTM

PR #1416, opened 2018-09-30
#630 - Add MolNet loaders for UV/Kinase/Factors datasets

VIGS25:
Dealing with #630 the PR adds in:

1. Loader functions for UV, Kinase and factors datasets.
2. UV tasks to separate file.
3. Loader functions inclusion in the run_benchmark.py file and the init files.


lilleswing:
Is this still WIP?

VIGS25:
I had a couple of questions, otherwise this is done.

1. For the UV dataset, the paper mentions 190 tasks, but the dataset contains tasks from 210-400 (both
inclusive) and that makes it 191 tasks. There's also a log TIC, but I am not sure if that is a task.
For now, I have added them all.

2. Was about to write tests, but I couldn't find any for loading the datasets, just the one for test_molnet.

3. Also, I wasn't sure whether to add any transformers. The kaggle_datasets.py had the transformers commented out.

VIGS25:
No, there are no other changes to make.
I had tested the loading myself, and it worked. So all good.

PR #1429, opened 2018-10-05
[WIP] #412: BEDROC metric

VIGS25:
Addressing #412, this PR adds in:

1. Implementation of the BEDROC metric

Tasks Left/Questions:
1. An additional argument of `alpha` needs to be supplied, which makes the API different from the rest of the metrics used. Need to fix that.
2. Tests for the metric

### DIFF
--- a/deepchem/metrics/__init__.py
+++ b/deepchem/metrics/__init__.py
@@ -161,6 +161,59 @@ def kappa_score(y_true, y_pred):
   return kappa
 
 
+def bedroc_score(y_true, y_pred, alpha=20.0):
+  """BEDROC metric implemented according to Truchon and Bayley that modifies
+  the ROC score by allowing for a factor of early recognition
+
+    References:
+      The original paper by Truchon et al. is located at
+      https://pubs.acs.org/doi/pdf/10.1021/ci600426e
+
+    Args:
+      y_true (array_like):
+        Binary class labels. 1 for positive class, 0 otherwise
+      y_pred (array_like):
+        Predicted labels
+      alpha (float), default 20.0:
+        Early recognition parameter
+
+    Returns:
+      float: Value in [0, 1] that indicates the degree of early recognition
+
+  """
+
+  assert len(y_true) == len(y_pred), 'Number of examples do not match'
+
+  assert np.array_equal(
+      np.unique(y_true).astype(int),
+      [0, 1]), ('Class labels must be binary: %s' % np.unique(y_true))
+
+  # Calculate ratio of actives to inactives
+  N = len(y_true)
+  n = sum(y_true == 1)
+  R_a = n / N
+
+  # The expression for the rie_denominator
+  rie_denom = R_a * (1 - np.exp(-alpha)) / (np.exp(alpha / N) - 1)
+
+  # Rank orders and rie_numerator
+  order = np.argsort(y_pred)
+  r_i = (y_true[order] == 1).nonzero()[0]
+  rie_numerator = np.sum(np.exp(-alpha * r_i / N))
+
+  # Rie_score
+  rie_score = rie_numerator / rie_denom
+
+  # Factor to be multipled
+  factor = R_a * np.sinh(
+      alpha / 2) / (np.cosh(alpha / 2) - np.cosh(alpha / 2 - (alpha * R_a)))
+
+  bedroc_score = rie_score * factor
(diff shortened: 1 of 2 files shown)


deepchem/metrics/__init__.py
```diff
@@ -161,6 +161,59 @@ def kappa_score(y_true, y_pred):
   return kappa
 
 
+def bedroc_score(y_true, y_pred, alpha=20.0):
```
lilleswing: Can we use the Bedroc code already in rdkit?
http://www.rdkit.org/Python_Docs/rdkit.ML.Scoring.Scoring-pysrc.html#CalcBEDROC

{code}
from rdkit.ML.Scoring.Scoring import CalcBEDROC
{code}

{code}
y_true = np.flatten(y_true)
y_pred = np.flatten(y_pred[:,1])
scores = list(zip(y_true, y_pred))
return CalcBEDROC(scores, 0, alpha)
{code}

VIGS25: The source for CalcBEDROC is a little confusing.

From the paper, what I gauged the idea should be is:
1. Sort the predicted scores/labels in increasing order and get the sorted indices.
2. Get ranks for active molecules by selecting the active molecule numbers from the sorted indices
3. Compute the expression for RIE.

What CalcBEDROC does:
1. We specify scores (which according to their unittest is a list of single element lists)
2. They directly check on the scores and compute number of actives and the RIE from that.

From the above code, the confusion stems from the fact that, if we supply scores as a list of (y_true, y_pred) pairs, the calcBEDROC never looks at the predictions to compute the RIE, which is weird.

If I use the reverse (ie (y_pred, y_true)) pairs, then the number of actives are directly computed from the y_pred values, which is again weird, because of presence of False Positives and Negatives in the prediction.

lilleswing: Ah I think we need to sort before calling CalcBedroc.

```python
from rdkit.ML.Scoring.Scoring import CalcBEDROC
y_true = np.flatten(y_true)
y_pred = np.flatten(y_pred[:,1])
scores = list(zip(y_true, y_pred))
scores = sorted(scores(scores, key=lambda x: x[1], reverse=True) # High is good
return CalcBEDROC(scores, 0, alpha)
```

If we sort first the RDKit code seems to be doing the algorithm laid out by you.

lilleswing: If I were to rewrite the docstring for CalcBedRoc after reading the source code.

```python
def CalcBEDROC(scores, col, alpha):
    """BEDROC original defined here: 
        Truchon, J. & Bayly, C.I. 
        Evaluating Virtual Screening Methods: Good and Bad Metric for the "Early Recognition" 
        Problem. J. Chem. Inf. Model. 47, 488-508 (2007). 

        scores: 2d sequence
             scores are a 2d sequence with the 0th index representing sample
             scores[sample_id] = vector of sample data
        col: int
             Index of sample data which reflects true label of a sample
             scores[sample_id][col] = True iff that sample is active
        alpha: float
             hyper parameter from the initial paper for how much to enrich the top

        
    """
```

lilleswing:
https://github.com/rdkit/rdkit/pull/2103

deepchem/metrics/__init__.py
VIGS25: Ah, I just checked their Github page 
https://github.com/rdkit/rdkit/blob/master/rdkit/ML/Scoring/Scoring.py

It says:
```scores: ordered list with descending similarity containing active/inactive information```

File: deepchem/metrics/__init__.py [hunk 2]
```diff
@@ -161,6 +161,47 @@ def kappa_score(y_true, y_pred):
...
+
+  """
+
+  assert len(y_true) == len(y_pred), 'Number of examples do not match'
+
+  assert np.array_equal(
+      np.unique(y_true).astype(int),
+      [0, 1]), ('Class labels must be binary: %s' % np.unique(y_true))
+
+  from rdkit.ML.Scoring.Scoring import CalcBEDROC
```
lilleswing: you can import this at the top -- no reason to do an in-function import.

peastman: Putting the import inside the function is a good idea, since we eventually want to make RDKit into a soft dependency.  See #1350.  It's going to take some work to remove all top level imports of RDKit, but for new code, let's go ahead and do it right.

lilleswing:
LGTM

PR #1484, opened 2019-01-06
Remove Six Dependency TextCNN

VIGS25:
Addressing @peastman's comments in #1478, `six` dependency has been removed.


peastman:
Looks great.

PR #1497, opened 2019-01-12
Use binding pockets for load_pdbbind

VIGS25:
Addressing #1404, this PR adds in the use of binding pocket for `load_pdbbind`.

@rbharath: `frag2_num_atoms` of 1000 works well, without any issues and the `AtomicConvModel` takes about 20 mins on CPU to train.


VIGS25:
Test failures was an unrelated stochastic one.

PR #1500, opened 2019-01-16
Improvements to SeqToSeq

peastman:
These are some minor changes that make SeqToSeq work a bit better when used as a VAE.


peastman:
The one test failure is an unrelated one in test_estimators.py.  This should be ready to merge.

Issue #1504, opened 2019-02-04
Symmetry functions featurization functions are not in "feats"


brunocuevas:
Hello

I just noticed that methods for featurization using Behler-Parrinello (2007) symmetry functions are not implemented in the corresponding "feat" class. On the contrary, those features are obtained using methods from deepchem.models.tensorgraph.symmetry_functions. It was a little hard for me to understand how to get those features until I read the commit of this class. 

I would suggest that the methods involved in the calculation of symmetry functions become methods from the deepchem.feat.BPSymmetryFunction, or on the contrary, if that is not possible, some reference to those functions in the docstrings.

Thanks, everybody for your great work

peastman:
The BPSymmetryFunction class is totally misnamed, because it doesn't do anything like what it claims.  It just creates an array of the coordinates and atomic numbers.  We should fix that.

For a lot of the common uses of these models (molecular dynamics, energy minimization), it's important that the inputs be raw coordinates and that the symmetry functions get calculated as part of the model.  That way, you can backpropagate through the whole thing and get derivatives of the energy with respect to the coordinates.

Do you have specific use cases for which you need the symmetry functions to be part of the featurization?

brunocuevas:
Yes, I wanted to predict the parameters of a model of electron density I'm working on, and I liked symmetry functions because they can provide specific atom feats. When I saw that there was a BPSymmetry class I thought I would have it coded in 5 minutes, but then I realized that is way more complex, and I don't know yet if any of the implemented models can fulfill my specific requirements.

peastman:
Can you explain what it is about your model that requires the inputs to be precomputed symmetry functions, rather than the atomic coordinates from which the symmetry functions are computed?

brunocuevas:
Actually, I must confess that I'm a newbie of deep learning, so at the beginning, I thought I'd just obtain a bunch of vectors, I'd store them in a matrix and then just train some NN. Of course, it was a way less sophisticated idea than including the symmetry functions as layers of the NN because, but it was good for me because I don't need to backpropagate on coordinates. After all, I'm trying to reproduce electron density, not energy. 

Anyway, I think I can try to use the classes of "deepchem.models.tensorgraph.symmetry_functions", it's just that it does not work the way I thought at the beginning when I read the name of this class

peastman:
[...]

PR #1513, opened 2019-02-19
#1504: Symmetry functions featurization functions are not in "feats"

VIGS25:
Addressing #1504, the function `BPSymmetryFunction` in `dc.feat.coulomb_matrices` has been renamed to `BPSymmetryFunctionInput`, since all the featurizer does is prepare the input matrices for the Symmetry Function layers.

Function name in all associated places has been changed.


VIGS25:
Unrelated stochastic test failure.

PR #1550, opened 2019-03-28
Remove uses of deprecated APIs

peastman:
This removes a lot of uses of deprecated APIs that will be going away in TensorFlow 2.  Lots more still to do, but it will be easier if I split it between several PRs.  This is the first batch.


peastman:
Some tests are failing because the signature for `tf.nn.dropout` changed between 1.12 and 1.13.  What do you think about bumping the development branch to TF 1.13, so we can use the non-deprecated argument?

VIGS25:
I would like the bumping to TF 1.13. I have suggested the same in my GSOC application, because of availability of a particular functionality of keras, only with TF 1.13.

peastman:
This should be ready to merge.

PR #1557, opened 2019-03-31
For DiskDataset: fix bug of move(); make output of create_dataset() clear

mapleleaf-soar:
**It fixs bug of DiskDataset.move() and makes output of DiskDataset.create_dataset() clear.**

I felt confused when using `load_pdbbind()` of `deepchem.molnet.load_function.pdbbind_datasets.py`.

### For addition in line 565

When using load funtion first time, I saw the output above:

> Featurizing Complexes
TIMING: dataset construction took 3.837 s
Loading dataset from disk.
Featurization complete.
TIMING: dataset construction took 3.738 s
Loading dataset from disk.
TIMING: dataset construction took 1.257 s
Loading dataset from disk.
TIMING: dataset construction took 2.201 s
Loading dataset from disk.

I felt strange because there were dataset already saved in my disk but it was my first time to use the load funtion. So I want to add this message to make it clear.

`Writing just created dataset to temp files of disk...`

### For modification in method `move()`

It should move **all files** in the `self.data_dir` into `new_data_dir`, but it moves **the folder** `self.data_dir` into `new_data_dir`.

> shutil.move(src, dst, copy_function=copy2)
Recursively move a file or directory (src) to another location (dst) and return the destination.
> If the destination is an existing directory, then src is moved inside that directory.

This will cause some problems, one of which is the newly generated dataset can **not** overwrite the previously saved dataset.

### DIFF
--- a/deepchem/data/datasets.py
+++ b/deepchem/data/datasets.py
@@ -562,6 +562,7 @@ class DiskDataset(Dataset):
 
     metadata_rows = []
     time1 = time.time()
+    print("Writing just created dataset to temp files of disk...")
     for shard_num, (X, y, w, ids) in enumerate(shard_generator):
       basename = "shard-%d" % shard_num
       metadata_rows.append(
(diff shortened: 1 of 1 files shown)


mapleleaf-soar:
image

For an instance of old `DiskDataset.move()`.

mapleleaf-soar:
@rbharath Any comments?

rbharath:
@peastman Would you mind taking a look? This touches the core `Dataset` class so I want to be careful before merging in.

peastman:
I'm a bit unclear on the reasons for this change.  We generally assume a 1-to-1 correspondence between models and directories on disk.  If I tell it to move the model directory to a new location, but there's already a directory there with that name, I'd expect it to throw an exception, because it probably means I made a mistake.  If I understand your change correctly, it would just merge the two directories together, combining all the files for both models in a single directory, which will probably produce a broken model directory.

Can you explain what problem you're trying to address?

PR #1566, opened 2019-04-05
Replaced Saver with Checkpoint

peastman:
The `Saver` class is going away in TensorFlow 2, so this replaces all uses of it with `Checkpoint`.  This is supposed to be backward compatible, so you *should* still be able to load checkpoints created with earlier versions of DeepChem, but please keep an eye out for any cases where that doesn't work.


peastman:
This should be ready to merge.

PR #1567, opened 2019-04-08
Replaced uses of deprecated layers

peastman:
`tf.layers` is going away in TensorFlow 2, so I've replaced all uses of that module with the corresponding Keras layers.  I also changed how layer variables are managed to better match how Keras handles them.


peastman:
The failure is in a test for ANIRegression's `save_numpy()` and `load_numpy()` functions.  Why does this model have its own mechanism for saving parameters that's different from the one used in all other models?  The mechanism it's using appears to be incompatible with Keras.

peastman:
@proteneer it looks like you created those functions.  Can I get rid of them and just have that class use the standard Checkpoint based mechanism for saving parameters?

proteneer:
It’s okay with me but I’d check with @lilleswing as well

peastman:
Is that OK with you Karl?

peastman:
I haven't gotten any reply from Karl, so I'm going to go ahead and remove those functions.  If there really is a reason that model needs its own save mechanism, it will need to be completely reimplemented in a different way.

PR #1587, opened 2019-05-02
Converted multitask models to KerasModel

peastman:
This is the first piece of converting the standard models to use KerasModel.  In addition to checking test cases, I also ran some of the examples to make sure they still work.  Almost everything was fine.  The one change I had to make was removing some totally unnecessary calls to `save()`.  I think those were left over from the pre-TensorGraph days, when `fit()` didn't automatically save the model parameters.

Please look this over carefully.  Once this is merged, people will be using KerasModel for real even if they don't know it!  Also be aware this will likely break compatibility with old checkpoint files (though I haven't tested that to be sure).

### DIFF
--- a/deepchem/hyper/grid_search.py
+++ b/deepchem/hyper/grid_search.py
@@ -77,8 +77,7 @@ class HyperparamOpt(object):
         model_dir = tempfile.mkdtemp()
 
       model = self.model_class(model_params, model_dir)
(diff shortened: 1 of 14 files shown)


peastman:
All tests are now passing.

peastman:
@rbharath have you had a chance to look at this yet?

PR #1595, opened 2019-05-14
Construct dataset first time, even with reload set to True

VIGS25:
This PR allows for constructing the featurized pdbbind dataset for the first time, even with reload=True. 
This is based on the idea that anyone running the code would do so multiple times and with reload=True, can save unnecessary computations and also not run into an error when the featurized datasets are not saved to disk initially.


VIGS25:
The test failure in Travis is an unrelated one.

PR #1600, opened 2019-05-18
Swapping split and transform order 

VIGS25:
I started to fix the order of train-valid-test split and the transformations on the data and I realized that the transformers would need to be separate for different datasets (as transformers take a dataset as argument). 

If a transformer takes a `train_dataset` as input, it would compute some statistics from that, which are later used to transform the valid and test, this would introduce dependencies between valid and train datasets and hence underestimate any downstream errors.

The current API of dc.load_xxx would then change to return transformers as a list of lists instead of list as returned now. ~~To the best of my knowledge, transformers are not reused explicitly anywhere after they are returned by the `load_XXX` function and hence changing the API a little will not be breaking, and even if it does, it would be minor.~~

A sample is shown for the `load_bace_classification`, and if alright with you guys, I will fix it similarly for all datasets.

Edit: `model.evaluate` uses transformers, for a dataset. Not sure whether to attempt to fix that, and break a lot of other things or just leave this situation as is

Thoughts - @rbharath & @peastman ?

### DIFF
--- a/deepchem/molnet/load_function/bace_datasets.py
+++ b/deepchem/molnet/load_function/bace_datasets.py
@@ -1,5 +1,5 @@
 """
-bace dataset loader.
+Bace dataset loader.
 """
 from __future__ import division
 from __future__ import unicode_literals
@@ -11,6 +11,19 @@ from deepchem.molnet.load_function.bace_features import bace_user_specified_feat
 
 logger = logging.getLogger(__name__)
 
+def get_transformers(dataset, mode="regression", move_mean=True):
+    """Returns transformers for each dataset, depending on task type."""
+    if mode == "regression":
+        transformers = [
+            deepchem.trans.NormalizationTransformer(
+                transform_y=True, dataset=dataset, move_mean=move_mean)
+        ]
+    else:
+        transformers = [
+            deepchem.trans.BalancingTransformer(transform_w=True, dataset=dataset)
+        ]
+
+    return transformers
 
 def load_bace_regression(featurizer='ECFP',
                          split='random',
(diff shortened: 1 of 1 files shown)


VIGS25:
Ah I see one test `TestGraphConvModel`, that uses the transformer returned from `load_function`, and throws up an error.

peastman:
If we're considering changing the API, what about having the Dataset directly keep track of Transformers that have been applied to it?  This would be a bigger change, but it might also be cleaner.  Loading functions wouldn't need to return a list of Transformers, and `predict()` and `evaluate()` could get them from the Dataset so you wouldn't need to pass them in.  We should think carefully about the consequences of this change, though.  If Transformers were used when creating a Dataset, would we always want to apply them during prediction?  Are there times when you want it to apply additional Transformers that weren't used in creating it?

VIGS25:
I would think this is a cleaner approach.

> If Transformers were used when creating a Dataset, would we always want to apply them during prediction?

If we are evaluating on a new dataset, different from the train-valid-test returned by the `load_*` method, then its better to apply the transformations while creating the dataset, and let the `predict()` or `evaluate()` function just run batches of the transformed dataset. 

> Are there times when you want it to apply additional Transformers that weren't used in creating it?

Do you have an example for this? - I am confused as to if it is ever done in a normal machine learning pipeline.

VIGS25:
Thought about this - Realised I had a mistake in my thinking. The test dataset is also transformed with respect to the train statistics, so that they are on the same scale as what the model as learnt from. So we need only one transformer. (https://sebastianraschka.com/faq/docs/scale-training-test.html)

Closing this issue now, will reopen again with the others fixed as well.

PR #1612, opened 2019-06-03
[WIP]: PretrainedModel support for KerasModel

VIGS25:
This is still a work in progress, but one that would make loading pretrained models with DeepChem much easier. The underlying code is the same as the `KerasModel`, but there are some key differences:

1. Instead of using `CheckpointManager`, I use `tf.train.Saver` to save and restore variables, as it takes an additional `var_list` argument, which allows you to specify scopes that you want to restore from. This makes it easier to do Transfer Learning. For example, I can train a model on ChEMBL dataset, and then just remove the Dense layers, while loading the pretrained model. The current `load_pretrained` function as an example does that. I would augment this by allowing certain scopes to be restored.

Another advantage of using `tf.train.Saver` over the `tf.train.CheckpointManager` is that of variable naming. There is the graph data and other metadata saved when using `Saver`, that allows me to specify variable names as is in the model and load them without any issues. When I tried using the same idea to the existing save mechanism, the names in the checkpoint were very different from the Tensor names. 

2. Using `tf.train.Saver` requires the full graph and ops to be built (I think there is a workaround around this, but its more complex), so this required me to move around the ops ordering a little.

All this code would go under `KerasModel` eventually, but for now, I made it into a different file, so as to not interfere with Peter's current work.

I ran this locally, and it works, using a Tox21 dataset for fine-tuning on the pretrained version of the ChEMBL small dataset.

@rbharath, @peastman: What are your views on this?


peastman:
I just recently put in a lot of work to replace all uses of Saver with Checkpoint.  Saver is going away in TensorFlow 2.0, and it doesn't work correctly with Keras models because they don't use consistent names for variables.  And it's generally a much less robust way of identifying how to save and reload variables.

VIGS25:
Ah I see. I agree the object based checkpointing that `Checkpoint` does is more robust, but the variable name based checkpointing that was done by Saver was much amenable to restore. 

Considering TF 2.0 is taking away Saver, no point continuing with this idea then.

VIGS25:
There's an easier way to do the loading using the `tf.keras.models.save_model` and `tf.keras.models.load_model`. Can that be used instead?

peastman:
Yes, that's the intended way of saving a Keras model.  See https://github.com/deepchem/deepchem/pull/1583#issuecomment-488118298.  Note that `save_model()` is different from a checkpoint.  It stores a complete representation of the model structure, not just the values of trainable variables.

VIGS25:
@peastman: Is this okay?
Both the checkpoint and keras based saving are present now, and `load_pretrained()` will not copy over weights from layers whose names are in `layers_to_exclude` argument.

VIGS25:
Those are some strange Travis failures, especially with variables not being initialized. I print out the list of variables initialized in every step, and the variables which throw up the not-initialized error, don't even show up.

peastman:
`load_pretrained()` isn't actually getting called in those test cases is it?  The only thing you've really changed is adding calls to `save_model()` in the training loop?  That must somehow be altering the TensorFlow graph.

In any case, I'm not sure it's a good idea to be calling `save_model()` every time you write a checkpoint.  That's going to add a lot of work and slow things down.  `save_model()` writes two types of information: the parameters, which have already just been written to the checkpoint, and the model topology, which doesn't change during training so there's no need to keep rewriting it.

Here are a couple of other issues I see.

1. You're trying to identify layers by name.  That isn't reliable, because layer names aren't consistent.  If you run the exact same code twice to create two identical models, their layers will have different names.

2. Keras models can be defined in two ways: by building them out of layers, or by subclassing Model  and implementing `call()`.  KerasModel needs to work with both kinds of models.

VIGS25:
No `load_pretrained()` is not getting called in these cases. `save_model()` is the only thing I added. It works with the `GraphConvModel`, and also `Smiles2Vec`, but I'm not sure why it fails in the case of `DTNNGraphModel`. 

Yes, I agree. I was not going to have both of them in the final version, but it looks as though `save_model()` is creating more intended errors.

Addressing your issues:
1. Yes, unfortunately. I couldn't find any relevant guide on how to restore a section of the model from object-based checkpoint, so I'm looking at TensorFlow's code to find some help for checkpoints. They have some utilities, but they also are driven by name of tensor or variable to restore.

If the create two identical models in the same script, I agree the naming will mess things over (In that case, two graphs need to be made and each model is managed by one graph)

2. Yup! 

Any other resource that you may know would also be helpful!

VIGS25:
Running `tf.train.list_variables()` on checkpoint file returns:
```
_CHECKPOINTABLE_OBJECT_GRAPH
model/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer-0/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer-6/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-0/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-0/embeddings/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-0/embeddings/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-0/embeddings/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-1/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/backward_layer/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/backward_layer/cell/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/backward_layer/cell/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/recurrent_kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/forward_layer/cell/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/forward_layer/cell/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/recurrent_kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-3/backward_layer/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-3/backward_layer/cell/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-3/backward_layer/cell/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/backward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/backward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/backward_layer/cell/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/backward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/backward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/backward_layer/cell/recurrent_kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/backward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/backward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-3/forward_layer/cell/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-3/forward_layer/cell/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/cell/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/cell/recurrent_kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-3/forward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-4/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-4/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-4/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-4/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-4/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-4/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-4/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
optimizer/beta1_power/.ATTRIBUTES/VARIABLE_VALUE
optimizer/beta2_power/.ATTRIBUTES/VARIABLE_VALUE
save_counter/.ATTRIBUTES/VARIABLE_VALUE
```

I could pass any of these names to `tf.train.load_variable(ckpt_file, variable_name)` which would give me the numpy value for the same  or  use `tf.train.init_from_checkpoint(ckpt_file, assignment_map)` that would replace the default initializer for the model with the checkpoint value. 

The task at hand now would be to not pass `model/layer_with_weights-4*` if one doesn't want the dense layers. For a small network this is easy to do, but if I want to fine-tune some more segments (which is the case with the ChemCeption model) doing this in an automated fashion would be tricky.

peastman:
Let's approach this from the user's perspective.  I believe this is the use case you're trying to support: you want to restore the model's parameters from a checkpoint, but you don't want to restore all of them.  You want to load only a subset of the parameters while leaving the others unchanged.  Is that correct?

What would be an ideal API for that feature?  How would you as a user want to specify which parameters to load and which to leave unchanged?

VIGS25:
Yes exactly, I don't want to load the last Dense Layer for example, since the number of tasks and task mode would be different between examples.

As a user, the simplest thing preferred is supply a name for the variables, and have those names restored. It would however be an issue since the naming is inconsistent, and calling the same Layer class multiple times in a script (even between different models in the same script) results in different names for equivalent layers between the models.

The next best solution would be to use layer numbers. I can also assume that the user knows which layer corresponds to what and supplying indices of the layers I don't want to restore should be sufficient. It also makes it easy with the current object based naming scheme to do a restore. Layer-based indexing is what Bert model also does, when trying to fetch embeddings.

peastman:
Neither of those approaches would work in general, especially for models defined by implementing `call()`.  What about instead specifying a list of variables?

VIGS25:
Even for the `call()` based models, the Checkpoint should be saved in the manner similar to the above example right?
If I specify `layers_to_exclude = [-1, -2]`, I would load everything except  `model/layer-with-weights-4/*`, `model/layer-with-weights-3/*`. These would be layers 5 and 4 in the original network. Since KerasModel needs a `tf.keras.Model` definition during initialization, calling `load_pretrained` after defining a KerasModel should work. 

This would look like:
```
model = KerasModel(args)
model.load_pretrained(ckpt_file, layers_to_exclude=[-1, -2])
```

I agree specifying a list of variables is the best option, but I am concerned that it is more applicable for someone who is comfortable with understanding how the organization inside the checkpoint is, but not sure if someone who just knows a little about transfer learning and wants to see it in practice.

peastman:
What does "layers" even mean in that case?  A Keras model doesn't need to be implemented with layers at all.

> I am concerned that it is more applicable for someone who is comfortable with understanding how the organization inside the checkpoint is

You wouldn't specify variables *in the checkpoint*, you would specify variables *in the model*.

VIGS25:
By Layers, I mean a grouping of some subset of model variables under a common identifer. 
For e.g, the kernels and biases under `Conv2D` would be grouped under `model/layer-with-weights-2/*`

When you say a Keras model can also not have any layers, is there an example you have? - Sorry I am confused on this,  or we could discuss this tomorrow during our DeepChem sync. I assumed that every model is made up of a chain of layers, each with its own position in the sequence.

Coming to the second point, 
If I print out `model.trainable_variables`, a subset of which is what you say we should specify, they look something like this:
```
<tf.Variable 'embedding/embeddings:0' shape=(54, 50) dtype=float32>
<tf.Variable 'Conv1D/kernel:0' shape=(3, 50, 192) dtype=float32>
<tf.Variable 'Conv1D/bias:0' shape=(192,) dtype=float32>
<tf.Variable 'bidirectional/forward_gru/kernel:0' shape=(192, 672) dtype=float32>
<tf.Variable 'bidirectional/forward_gru/recurrent_kernel:0' shape=(224, 672) dtype=float32>
<tf.Variable 'bidirectional/forward_gru/bias:0' shape=(672,) dtype=float32>
<tf.Variable 'bidirectional/backward_gru/kernel:0' shape=(192, 672) dtype=float32>
<tf.Variable 'bidirectional/backward_gru/recurrent_kernel:0' shape=(224, 672) dtype=float32>
<tf.Variable 'bidirectional/backward_gru/bias:0' shape=(672,) dtype=float32>
```

Inside the checkpoint, the equivalent would be:
```
_CHECKPOINTABLE_OBJECT_GRAPH
model/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer-0/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer-6/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-0/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-0/embeddings/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-0/embeddings/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-0/embeddings/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-1/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-1/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/backward_layer/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/backward_layer/cell/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/backward_layer/cell/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/recurrent_kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/backward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/forward_layer/cell/.ATTRIBUTES/OBJECT_CONFIG_JSON
model/layer_with_weights-2/forward_layer/cell/bias/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/bias/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/recurrent_kernel/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/m/.ATTRIBUTES/VARIABLE_VALUE
model/layer_with_weights-2/forward_layer/cell/recurrent_kernel/.OPTIMIZER_SLOT/optimizer/v/.ATTRIBUTES/VARIABLE_VALUE
optimizer/beta1_power/.ATTRIBUTES/VARIABLE_VALUE
optimizer/beta2_power/.ATTRIBUTES/VARIABLE_VALUE
save_counter/.ATTRIBUTES/VARIABLE_VALUE
```

Restoring those variables would involve an additional mapping step, with the only way of linking them being the names.
If I have two models definitions in my script, and use a `Bidirectional` layer, the names of the variables would be `bidirectional` and `bidirectional_1`. If my pretrained model had only the variable name `bidirectional`, which would mostly be the case (since its the only model definition in my script), then using this mapping would throw up an error for the second model.

However, if I use the position of the Bidirectional Layer in my model (Say something like Input -> Conv2D -> Bidirectional, with bidirectional having position 3), then I can just call the `model/layer-with-weights-3/*` for restoring my bidirectional variables for both models.

peastman:
> When you say a Keras model can also not have any layers, is there an example you have?

There's a discussion of the different ways of creating models at https://www.tensorflow.org/alpha/guide/keras/functional.  One way is to create a class that extends Model, allocate any variables you want in the constructor or `build()`, and then do any calculation you want inside `call()`.  It's possible that no Layer objects will ever be involved.

> For e.g, the kernels and biases under Conv2D would be grouped under model/layer-with-weights-2/*

You're still thinking in terms of the content of the checkpoint file.  Like you said, most users have no clue how checkpoint files are organized, and they shouldn't have to.  A friendly API should only involve things they already know about.  For example, suppose you want to restore only a single layer from the checkpoint.  A natural API would be

```python
model.restore(variables=layer.variables)
```

With one glance, you understand exactly what that call is doing because it only involves things you're familiar with.  If you want to make it restore the variables from multiple layers, you know how to do that too.  You don't need to learn about string identifiers or the hierarchical arrangement of variables inside files.

VIGS25:
I understand what you're trying to say and this is the kind of API one would ideally want. However, the utilities TensorFlow provides for dealing with checkpoints don't understand the Tensors present in `layer.variables`. 

They all expect variable names based on the hierarchical arrangement of variables inside files, or some mapping to do that (Link below). https://github.com/tensorflow/tensorflow/blob/873dee2ac969474c93ded53386d9b98d58b92031/tensorflow/contrib/framework/python/framework/checkpoint_utils.py#L152

peastman:
Let's first figure out what we want the API to be.  Then we can figure out how to implement it! 😀

VIGS25:
Haha :D, Okay!

Then having something like what you said should be perfect. We can these functions:
```
model.restore_layer(variables=layer.variables)
```
```
model.restore(layers=layers)
```
```
def restore(self, layers=None):
     for layer in layers:
           model.restore_layer(variables=layer.variables)
```

peastman:
Ok, now we just need to figure out how to make it happen!

Well, there's one trivial way of doing it.  Record the values of all variables in the model that we don't want to get loaded.  Go ahead and load the checkpoint.  Now set all the variables that shouldn't be changed back to the values we recorded.  Possibly there's a more elegant way of doing it.  But do you see any reasons this wouldn't work?

VIGS25:
Loading layer by layer from a checkpoint is actually hard to do right now, and becomes cumbersome for larger models. It makes more sense as you said to load the full checkpoint first and then copy over values.

One issue I see with this is that the model I load the weights into should have the same architecture as the pretrained one, an example being the differing number of units in the final Dense layer between the model I want to restore to, and the pretrained on.

One work-around this would be to have a dummy model (in a different session), that has the same architecture as the pretrained one, to which we load weights into, and then we copy over the layers or layer variables we want, into the original model.

peastman:
Let's go back to the API and clarify the intended use case.  I was assuming you wanted to load a subset of the variables for the *same model*.  But you're saying you want to load variables from a *different model* that has a different architecture?  How do you imagine that working (from a user's perspective)?  How would you tell it which model to load from, which model to load into, and how to match up variables between them?

VIGS25:
Eventually I want to be able to do both. 

### Expected functionality:

1. Restoring to a same architecture model - You will always load the checkpoint here. You can either continue fine-tuning or training from there or freeze some layers (can be done as simply in `tf.keras` as `layer.trainable = False`) and then do the fine-tuning. 

2. Restoring to a different architecture model: The simplest thing that can work for most deepchem models is that of classification vs regression and changes in Dense Layers. 

### Expected APIs:

For now, we can keep the APIs separate: (Names are weird but just for clarification)

1. Restore to a same architecture model: 
```
def restore_from_same(self, ckpt=None, layers_to_freeze=None):
      # Restore everything
      # Set layers in layers_to_freeze to False (Should recompile model after that)
```

2. Restore to a different architecture model:
We can start simple, allowing for changes in final Dense Layers, just like Keras currently offers. This makes it to implement as architectures would be the same until the final dense layer(s), so copying over would be easy. (A single pass through the layers should do).

```
def restore_from_other(self, ckpt=None, layers_to_freeze=None,  include_top=False):
      # First restore everything to dummy model.
      # Build Graph for new model and copy over layer weights to this, 
      # except for the last Dense Layer
      # if include_top is True then you can call restore_from_same
```

peastman:
> We can start simple, allowing for changes in final Dense Layers

What do you mean by "final Dense layers"?  You seem to be assuming some very specific architecture.  Remember:

- Keras models are not necessarily made of layers.
- When they are made of layers, those layers can be arranged in arbitrary topologies.
- Keras models are composable.  A model might include other models inside it.

> (Set layers in layers_to_freeze to False (Should recompile model after that)

I don't understand what this means.  What does it mean to "set a layer to False?"  Can you explain?

> if include_top is True then you can call restore_from_same

I don't understand what this means either.

VIGS25:
Sorry for the confusion, but I have been thinking only about a majority of *models DeepChem provides right now* for supporting Transfer Learning, which was the basis of my proposal for Google Summer of Code and ChemNet. So, all my assumptions about the architecture derive from that. `Seq2Seq` or `GANs` are not on my focus right now, since they are composed of multiple models.

A couple of examples:
https://github.com/deepchem/deepchem/blob/ff21bdac0f82311cde684e4f081ea5a40472806d/deepchem/models/tensorgraph/models/text_cnn.py#L143

https://github.com/deepchem/deepchem/blob/ff21bdac0f82311cde684e4f081ea5a40472806d/deepchem/models/tensorgraph/models/graph_models.py#L257

This is the Dense Layer I refer to, the one going from some features to the task specific output.
A Keras model may not consist of layers yes, but something for which an architecture is already built into DeepChem does. For models already defined, the topology will not be arbitrary right? 

>  (Set layers in layers_to_freeze to False (Should recompile model after that)

I'm sorry. What I meant to say was that `layer.trainable` attribute can be set to `False`.

> if include_top is True then you can call restore_from_same

`restore_from_same` is the API to load models with the same architecture as the pretrained one, and it is used for the same mode (classification or regression).

`restore_from_other` has an `include_top` argument (taken from the Keras API). This is a choice to allow the user to remove the Dense Layer in the end, which means they either want to load existing weights for a different mode (classification vs regression), or the number of tasks varies between the pretrained model and the one we load into. `include_top` set to True just means the pretrained and current architectures are same. The restore can then be done by using `restore_from_same`.

I agree this is not very general though, but atleast a start with building a Keras equivalent, since their pretrained support is also limited to removing `Dense` layers in the end. They have a lot of other convenience functions, but that needs some experience on the user side, with transfer learning.

peastman:
> For models already defined, the topology will not be arbitrary right?

Specific model subclasses could provide mechanisms like that.  But anything in `KerasModel` needs to be able to work with arbitrary instances of `tf.keras.Model`.  So let's think about the situation in more generic terms.

You have a trained model.  It has some topology and some set of parameters.

You also have a different model with a (possibly) different topology.  You haven't trained that model yet.  You want to copy over a subset of the parameters from the first model as a starting point for training.  Right?

Depending on the situation, the particular set of parameters you want to copy could vary.  The similarity between the architectures could also vary.  Here are some common examples.

- The architectures are basically identical, with the possible exception of the widths of the output layers.  You've trained a model on one set of tasks, and now you want to use that as a starting point to train a model with identical inputs on a different set of tasks.

- The earlier parts of the model are identical, but then they diverge.  An example would be training an image recognition model, then using it as the first half of a UNet for doing image transformations.  The intermediate representation learned by the original model is likely to be useful for the new task, but the second half of the new model has no relation to anything in the original model.  Another example would be training an autoencoder, then using the first half of it to generate the intermediate representation for a classification model.

- It's also possible that it's the later parts of the model you want to copy, not the earlier ones.  This could happen if you train an autoencoder, then want to use the decoder half as part of a generative model.

So as I see it, the real task we're trying to perform is copying parameters between models.  In its very simplest form, this feature might involve a method like

```python
copy_from(source_model, variable_map)
```

where `source_model` is the model you want to copy variables from, and `variable_map` is a dict mapping from variables in the source model to variables in the new model.  A more sophisticated form might try to work out the correspondences automatically by comparing the architectures of the two models.  That could save the user some work, though it also would have a risk of failing silently in unexpected ways (working out a different mapping than what the user wanted).

VIGS25:
Moving to #1643

PR #1615, opened 2019-06-04
Converted more models to Keras

peastman:


peastman:
All tests are passing.

PR #1633, opened 2019-07-02
[WIP]: Validation dataset logging and metrics

VIGS25:
First steps towards unifying metric and loss computation for validation datasets.

Major changes:
1. Gradient based ops for eager and graph mode moved to separate functions, and can handle both train and eval modes.

2. Create `loss_tensor`, `loss_fn` and `train_op` before training, as opposed to during training.

### DIFF
--- a/deepchem/models/keras_model.py
+++ b/deepchem/models/keras_model.py
@@ -207,7 +207,7 @@ class KerasModel(Model):
       self._loss_outputs = list(range(len(self._output_tensors)))
     self._init_new_vars()
 
-  def _create_training_ops(self, example_batch):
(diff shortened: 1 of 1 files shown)


peastman:
I don't think we're ready to start implementing this yet.  As I said in #1628, we need to consider the design carefully.  There are a lot of issues to consider, and they need to be discussed.

VIGS25:
What do you think the main issues would be?
From what I see, the ops would be the same for the datasets, except for gradients computed on the train dataset.

Apart from that, I have had issues wrapping wthe `train_op` and `loss_tensor` into a clean API, but that can be fixed.

peastman:
Let's discuss it under #1628.

PR #1638, opened 2019-07-14
ChemNet Fixes and Additions

VIGS25:
This PR adds in enhancements in line with the ChemNet paper:

**Transformers**:
1. MinMax transformer is added along with tests and documentation - This is the Transformer used in ChemNet pretraining. 

**Tox21, HIV, ChEMBL25, FreeSolv**
1. Train, Valid and test split fractions added in kwargs for all the four datasets
2. RandomStratifiedSplitter added in Tox21 and HIV, since that is used in the paper for original training and fine-tuning.
3. Custom directories for data download and dataset storage for all datasets
4. MinMax transformer has also been added to `dc.molnet.chembl25_datasets`

**Featurizers**:
1. Safeguard against the NaN problem in SmilesToImage featurizer

**KerasModel**
1. Replaced the `print()` statements for losses with logger statements. A large fraction of DeepChem uses loggers, so this would make it more compatible

2. Losses were getting logged only while saving checkpoints. In general, one might want to keep them separate, since the frequencies of monitoring losses vs saving can be different then.

3. Added support for custom directories when saving checkpoints, or restoring existing model, or getting a list of checkpoints.

### DIFF
--- a/deepchem/models/keras_model.py
+++ b/deepchem/models/keras_model.py
@@ -1,6 +1,10 @@
 import numpy as np
 import tensorflow as tf
 import time
+import logging
+import os
+
+logger = logging.getLogger(__name__)
 
 from deepchem.data import NumpyDataset
 from deepchem.models.losses import Loss
@@ -348,7 +352,6 @@ class KerasModel(Model):
       inputs, labels, weights = self._prepare_batch(batch)
       self._tensorboard_step += 1
       should_log = (
-          self.tensorboard and
           self._tensorboard_step % self.tensorboard_log_frequency == 0)
       if tf.executing_eagerly():
 
@@ -401,7 +404,7 @@ class KerasModel(Model):
                   loss_tensor, global_step=self._global_step, var_list=vars)
             train_op = self._custom_train_op[op_key]
         fetches = [train_op, self._loss_tensor, self._global_step]
-        if should_log:
+        if self.tensorboard and should_log:
           fetches.append(self._summary_ops)
         feed_dict = dict(zip(self._input_placeholders, inputs))
         feed_dict.update(dict(zip(self._label_placeholders, labels)))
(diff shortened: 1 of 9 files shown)


VIGS25:
Travis Failure is unrelated.

peastman:
> Replaced the print() statements for losses with logger statements.

I intentionally used print() here because the logger isn't working correctly, and we've never been able to figure out why.  See #1390.  If you can figure out how to fix that, we can switch over to logger.

VIGS25:
I think there was a fix proposed for this in #1591, and I ran my jobs with logger and it works properly.

deepchem/models/keras_model.py
```diff
@@ -922,12 +934,20 @@ def _exec_with_session(self, f):
...
+    model_dir: str, default None
+      Directory to get list of checkpoints from. Reverts to self.model_dir if None
+
+    """
+    if model_dir is None:
+      model_dir = self.model_dir
+    return tf.train.get_checkpoint_state(model_dir).all_model_checkpoint_paths
 
-  def restore(self, checkpoint=None):
+  def restore(self, checkpoint=None, restore_from=None):
```
peastman: The argument should be called `model_dir` for consistency with `save_checkpoint()` and `get_checkpoints()`.

peastman:
Ok, if it's working now that's great.

deepchem/models/keras_model.py
VIGS25: Done

deepchem/models/keras_model.py
```diff
@@ -936,10 +956,14 @@ def restore(self, checkpoint=None):
       the path to the checkpoint file to load.  If this is None, the most recent
       checkpoint will be chosen automatically.  Call get_checkpoints() to get a
       list of all available checkpoints.
+    restore_from: str, default None
```
peastman: It still has the old argument name here.

peastman:
One minor mistake in a docstring.  Otherwise it looks good.

deepchem/models/keras_model.py
VIGS25: Fixed.

PR #1643, opened 2019-07-18
First version of pretrained loading

VIGS25:
This is a starter version for transfer learning from the discussion in #1612.

It is expected that the user gives the correspondence between variables to restore from and variables to restore to. I could not find a way to make variables non-trainable, but @peastman has alllowed only a list of variables to be optimized during training, so that should work.

The future work would be to moving towards making it easier for users to do Transfer Learning. 
One idea would be to have a default setting, where only the last dense layer is replaced, depending on task type and number of tasks, and the rest of the layers are restored as is. For the more advanced users, the current option can be retained.

### DIFF
--- a/deepchem/models/keras_model.py
+++ b/deepchem/models/keras_model.py
@@ -11,6 +11,7 @@ from deepchem.models.losses import Loss
 from deepchem.models.models import Model
 from deepchem.models.tensorgraph.optimizers import Adam
 from deepchem.trans import undo_transforms
+from deepchem.metrics import to_one_hot
 from deepchem.utils.evaluate import GeneratorEvaluator
 
 
@@ -981,6 +982,21 @@ class KerasModel(Model):
       return int(self._global_step)
     return self._global_step.eval(session=self.session)
 
+  def compute_loss(self, dataset, transformers):
+    if self.mode == "regression":
+      y = undo_transforms(dataset.y, transformers=transformers)
+    else:
+      y = to_one_hot(dataset.y.flatten(), self.n_classes).reshape(
+          -1, len(self.n_tasks), self.n_classes)
+    loss_fn = model._loss_fn
+
+    y_pred = self.predict(dataset, transformers=transformers)
+    loss_tensor = loss_fn([y_pred], [y], [dataset.w])
+    if tf.executing_eagerly():
+      return loss_tensor.numpy()
+    else:
+      return self.session.run(loss_tensor)
+
   def _create_assignment_map(self, source_model, include_top=True, **kwargs):
     """
     Creates a default assignment map between variables of source and current model.
@@ -1098,12 +1114,10 @@ class KerasModel(Model):
         dest_var.assign(value_map[source_var])
 
     else:
-      self._assign_ops = []
       with self.session.as_default():
         for source_var, dest_var in assignment_map.items():
           assert source_var.shape == dest_var.shape
           assign_op = dest_var.assign(value_map[source_var])
-          self._assign_ops.append(assign_op)
           self.session.run(assign_op)
 
     dest_vars = list(assignment_map.values())
--- a/deepchem/models/keras_model.py
+++ b/deepchem/models/keras_model.py
@@ -953,6 +953,37 @@ class KerasModel(Model):
       return int(self._global_step)
     return self._global_step.eval(session=self.session)
 
+  def load_pretrained(self,
+                      assignment_map=None,
+                      checkpoint=None,
+                      model_dir=None):
+    """Load from a pretrained model.
+
+    Parameters
+    ----------
+    assignment_map: Dict, default None
+      Dictionary containing variable mapping between source and current model
+      variables
+    """
+    self._ensure_built()
+    if assignment_map is None:
+      self.restore(checkpoint=checkpoint, model_dir=model_dir)
+    else:
+      if tf.executing_eagerly():
+        for source_var, dest_var in assignment_map.items():
+          dest_var.assign(source_var)
+      else:
+        self._assign_ops = []
+        for source_var, dest_var in assignment_map.items():
+          assign_op = dest_var.assign(source_var)
+          self._assign_ops.append(assign_op)
+          self.session.run(assign_op)
+
+        if hasattr(self, '_initialized_vars'):
+          self._initialized_vars.update(set(assignment_map.values()))
+        else:
+          self._initialized_vars = set(assignment_map.values())
+
 
 class _StandardLoss(object):
   """The implements the loss function for models that use a dc.models.losses.Loss."""
--- a/deepchem/models/keras_model.py
+++ b/deepchem/models/keras_model.py
@@ -1010,9 +1010,34 @@ class KerasModel(Model):
 
     return assignment_map
 
+  def _create_value_map(self, source_model, **kwargs):
+    """
+    Creates a value map between variables in the source model and their
+    current values. This is used only when a custom value map is missing.
+
+    Parameters
+    ----------
+    source_model: dc.models.KerasModel
+        Source model to create value map from
+    include_top: bool, default True
+        if true, copies the last dense layer
+    """
+    value_map = {}
+    source_vars = source_model.model.trainable_variables
+
+    if tf.executing_eagerly():
+      for source_var in source_vars:
+        value_map[source_var] = source_var.numpy()
+    else:
+      for source_var in source_vars:
+        value_map[source_var] = source_var.eval(session=self.session)
+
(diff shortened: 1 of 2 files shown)


peastman:
This method seems like two unrelated methods combined together.  If `assignment_map` is None, it loads a model from disk.  In fact, in that case this method is just a synonym for calling `restore()`.  If `assignment_map` is not None, then it doesn't load anything, which makes the name misleading.  It just copies values between Variable objects.

VIGS25:
I understand and should fix it soon, but the idea was to either restore the whole model (that is copy all variables), or copy a section of the variables. I should rename it to `copy_from`.

A much cleaner way would be to just take in checkpoint or model directory, restore everything into a model and then copy over from there. All of this can be done inside this function. The only trouble is with constructing the assignment_map inside, or even being able to supply it as an argument.

VIGS25:
Moving towards a layer based assignment map and taking advantage of `tf.keras` utilities for loading and saving weights. 

Can switch back to the variable based version if needed.

deepchem/models/keras_model.py
```diff
@@ -969,14 +971,89 @@ def restore(self, checkpoint=None, model_dir=None):
...
+        session = self.session
+      self._checkpoint.restore(checkpoint).run_restore_ops(session)
 
   def get_global_step(self):
     """Get the number of steps of fitting that have been performed."""
     if tf.executing_eagerly():
       return int(self._global_step)
     return self._global_step.eval(session=self.session)
 
+  def default_assignment_map(self, source_model, include_top=True, **kwargs):
```
peastman: How about making this method private?  That is, rename it to `_default_assignment_map()`.  It's just there to be called from `load_from_pretrained()`, not for users to call directly.

deepchem/models/keras_model.py
```diff
@@ -969,14 +971,89 @@ def restore(self, checkpoint=None, model_dir=None):
...
+        if true, copies the last dense layer. Used only when assignment map is None
+    """
+    self._ensure_built()
+    if assignment_map is None:
+      assignment_map = self.default_assignment_map(
+          source_model=source_model, include_top=include_top)
+
+    self._assign_ops = []
+    if tf.executing_eagerly():
+      source_model.restore(model_dir=model_dir, checkpoint=checkpoint)
```
peastman: I'm a bit nervous about calling `restore()` here.  `source_model` is just supposed to be the place values are taken from.  You don't expect it to get modified by this call.  This could lead to a lot of surprises for users.

deepchem/models/keras_model.py
```diff
@@ -969,14 +971,89 @@ def restore(self, checkpoint=None, model_dir=None):
...
+    which assumes the architecture has a Dense layer as its last layer and can be
+    included by setting include_top to True. The default assignment map allows
+    applying an existing trained model to a different multi-task setting involving
+    the same model or to a different task like classification from regression or
+    vice-versa.
+
+    Parameters
+    ----------
+    assignment_map: Dict, default None
+      Dictionary containing layer mapping between source and current model layers
```
peastman: I think this should be a map between variables, not layers.  Remember, not all models are composed of layers!

deepchem/models/keras_model.py
VIGS25: While the current API does not reflect that, with the `default_assignment_map`, I was aiming for something like `default_generator`. But I am not sure if it is better for someone to change `default_assignment_map` or supply an `assignment_map`. I think the latter is the preferred option and in that setting it should be made private.

deepchem/models/keras_model.py
VIGS25: `source_model` is just a container to copy values into, like a dummy and the idea was to have a checkpoint to restore from. But I could add a toggle that allows users to restore if they want to, else just use the current values of a model trained in the same script as finetuning.

deepchem/models/keras_model.py
VIGS25: `KerasModel` can be built in two ways:
1. Using the functional API which is a DAG of layers. This would obviously work with the current setup.

2. Subclassing `tf.keras.Model` and implementing `__call__()`. For the second one, all examples are defined similar to the example below. The layers are defined at some point in the `__init__` or `build()` and are added to the `layers` attribute of the model. 
```
class MLP(tf.keras.Model):

  def __init__(self, **kwargs):
    super(MLP, self).__init__(**kwargs)
    self.dense_1 = layers.Dense(64, activation='relu')
    self.dense_2 = layers.Dense(10)

  def call(self, inputs):
    x = self.dense_1(inputs)
    return self.dense_2(x)

example_model = MLP()
print(example_model.layers)
```
The output has both the dense layers defined. Same thing happens when used with `build`. However if the variables or layers are defined under `call`, then this won't work. But is that usual practice?

I can switch back to variables, but its easier to keep track of layers (and associated variables indirectly) because there are fewer of them when compared to variables.

peastman: It doesn't even need to use layers at all.  It can define variables directly in `__init__()` or `build()`, then make use of them in `call()`.  This style of model definition was added recently.  It closely matches the way models are defined in PyTorch, so I'd expect it to be pretty common going forward.

Another option is that we could allow the map to contain either layers or variables.  I'm not sure that really adds much, though, other than being a potential source of confusion.

deepchem/models/keras_model.py
VIGS25: This is done.

deepchem/models/keras_model.py
VIGS25: I realize that `restore` needs to be the there because the pretrained and fine-tuned model run under different Sessions, and the value of the same variables are different between sessions. `tf.assign` is not capable of copying values from one session to another. In this case, values need to be written to disk and hence `restore` is needed. I will mention this explicitly in the docs though.

peastman: So retrieve the values of the variables as numpy arrays and then set them in the other session.

VIGS25: Done. Sorry I had a couple of exams this week, and got to it now.

peastman:
What is `self._assign_ops` for?  You add ops to it, but I don't see it ever getting used.

VIGS25:
Nothing special. Just a container to keep track of the ops used for assignment.

peastman:
I mean, to keep track of them for what?  You never use them again.  Just curious.

VIGS25:
In the initial stages, I had that just to check what were the the `assign_ops` that were executed and if all the expected ones were there. Now it serves no purpose in there, I will remove it.

VIGS25:
All tests pass now. This is ready to be merged. @peastman: Any other suggestions?

peastman:
It's looking good!  Just a few notes on incorrect docstrings:

- `_create_value_map()` documents an argument called `include_top` even though it actually has no such argument.
- `load_from_pretrained()` incorrectly describes `assignment_map` as "Dictionary containing layer mapping between source and current model layers".  It's actually a map between variables, not layers.
- It also incorrectly says, "source_model is a dc.KerasModel with the same architecture as the pretrained model."  `source_model` *is* the pretrained model.  Did you mean to say, "the same architecture as this model"?

A couple of notes on the test cases.  Rather than duplicating all the code for each test, once for graph mode and once for eager mode, you can just have one call the other:

```python
def test_load_from_pretrained_eager_mode(self):
    with context.eager_mode():
        self.test_load_from_pretrained_graph_mode()
```

When constructing a model in a test case, don't specify a value for `model_dir`.  By doing that, you cause it to be saved in a regular directory on disk that gets left in the middle of your source tree.  If you omit `model_dir`, it uses a temporary directory that gets deleted automatically.

Are you sure the following code is correct?

```python
    source_vars = source_model.model.trainable_variables
    dest_vars = self.model.trainable_variables

    if not include_top:
      source_vars = source_vars[:-2]
      dest_vars = dest_vars[:-2]
```

Does Keras make any guarantees about the order of the variables in `trainable_variables`?  Is it really certain the last two are from the final layer?

VIGS25:
Addressing incorrect docstrings:
1. Removed `include_top` for `_create_value_map()`
2. Fixed the docstring for `assignment_map`
3. So this was the case when I was always restoring from the checkpoint. A different instance with the same configuration as the pretrained model was created and that was `source_model`. Now with the use of `value_map`, this can either be the pretrained model or a copy. The docstring has been updated to reflect that.

Fixed the issues with eager vs graph mode tests.

Yes, in Keras, the tensors are added to `trainable_variables` in the order of creation, so the last two are from the final layer.

peastman:
Yes, I think it's ready to merge!

Issue #1644, opened 2019-07-19
Upgrade to TF 1.14


VIGS25:
The developers at TensorFlow addressed #1642, saying that this issue has been fixed in TF 1.14, and indeed it has been.

After upgrading to TF 1.14 on a local branch and running tests - most tests passed. The ones which were failing have been fixed except for one, which tries to fit a GAN with multiple generators and discriminators.

I isolated the reason for the failure, and this is a Keras issue and not a DeepChem one.

TF 1.13.1
https://github.com/tensorflow/tensorflow/blob/6612da89516247503f03ef76e974b51a434fb52e/tensorflow/python/keras/engine/network.py#L960

TF 1.14
https://github.com/tensorflow/tensorflow/blob/a5120db4d6917f943176ef3c5bb938064604c761/tensorflow/python/keras/engine/network.py#L849

If you see, there's an extra line `depth_keys = depth_keys[1:]` which was not there in TF 1.13.1
This is preventing some tensors from being computed and hence throwing up an issue with multiple generator-discriminator GAN.

If we should upgrade to TF 1.14, then for now, the GAN test can be marked with `attr('slow')`.

peastman:
I don't think this looks like a TensorFlow bug.  It's an intentional change in behavior.  So we need to understand what the change is, why they made it, and what assumptions it's based on.  Then we need to make the appropriate changes to our code to work correctly with the new behavior.

Possibly we'll need to check the TF version at runtime and use a different behavior depending on whether the version is <1.14.  Hopefully that won't be necessary, but we can do it if needed.

VIGS25:
The comment in the line of change says, "Ignoring input layers while building the graph". 

L139-144 in  `dc.models.tensorgraph.models.gan` that are getting affected by this change and the tensors corresponding to them are not computed.
 
https://github.com/deepchem/deepchem/blob/94809d39b5cee0c1ba178a1936dbf9a95b5eaad7/deepchem/models/tensorgraph/models/gan.py#L139

How does one go about generally trying to fix these things?

peastman:
I'm looking into it now.  How did you determine that particular line is responsible?

VIGS25:
The Lambda layer corresponding to L153 was not getting computed, which was the error thrown. It takes in two layer outputs - `stacked_gen_loss` & `weight_products`. 

I track the tensors that are computed while the graph is built. `stacked_gen_loss` is computed initially, `weight_products` is missing, probably because of the change in line of code between the versions.

[...]

PR #1645, opened 2019-07-19
Upgrade to TF 1.14 (Optional)

VIGS25:
This PR makes the necessary changes for upgrade to TF 1.14. 

Main Changes:
1. Remove extra addition of variables, which are already part of `layer.trainable_variables`
2. Install script change
3. Made `attr('slow')` for failing GAN test (Issue on Keras front, discussed in #1644)
4. Moved activations.get to __init__, fixes another error.


peastman:
We should fix the GAN failure as part of this PR.  Masking a failing test isn't what `slow` is for.

VIGS25:
I would want to, but I think the issue is on the TensorFlow's end.

VIGS25:
Do we merge this with the current fix, and later remove the `attr('slow')` after TF fixes the bug?

peastman:
No, we shouldn't merge this until we've resolved the issue one way or another.

peastman:
And everything is passing.  Yay!

VIGS25:
So this can be merged now?

peastman:
Looks good to me.  I'll leave it to @rbharath to merge, since he hasn't commented on it yet.

PR #1651, opened 2019-08-03
ChemNet Fixes

VIGS25:
Couple of minor fixes to ChemNet.
1. Changed the 2-class classification case to use SigmoidCrossEntropy loss
2. SingletaskStratifiedSplitter added for `sampl` dataset.


VIGS25:
Travis failure is only in Python 2.7 and on an unrelated test.

PR #1658, opened 2019-08-16
Moved to Python 3.5 and 3.7 for Travis

peastman:


peastman:
That took more work than I expected.  I ended up modifying the install script to not specify particular library versions for anything except TensorFlow.  Before we were specifying particular versions for most of them, and that led to a mess of version conflicts when installing on Python 3.7.

Once I did that, it revealed another problem.  We were using the Scipy functions for reading and writing images.  Those don't exist anymore in the latest version of Scipy, so I converted them to use PIL instead.

There also were some test failures caused by a bug that was introduced in #1649.  I fixed that.

Everything is now passing, so it's ready to merge if it looks ok.

peastman:
@rbharath is this ok to merge?

VIGS25:
I fixed the bug too, in #1660.

PR #1664, opened 2019-08-22
Reorganized models directory

peastman:
I'm trying to figure out why #1656 keeps failing on travis.  So I'm going to try adding small pieces of it to a new PR one at a time.


peastman:
I've now made most of the same changes here that are in #1656, and yet it's still working.  So I'm going to go ahead and merge this in place of that one, then continue to investigate the final changes.

PR #1666, opened 2019-08-23
Move test cases out of tensorgraph module

peastman:
More changes from #1656.


peastman:
Apparently moving the test cases is what causes the failures.  Travis reports lots of memory errors.  Not in any of the tests that got moved, of course.

peastman:
Apparently the errors are caused by moving test_seqtoseq.py.  Why?  Who knows.  Anyway, I'll move the other tests.

PR #1686, opened 2019-09-26
Support residual networks

peastman:
Implements #1681.  So far I've implemented it for MultitaskClassifier and MultitaskRegressor.  I'm posting this now to get feedback.


peastman:
I think we might as well merge this.  I was originally planning to also support this feature in all the graph models.  But when I started looking at them, I discovered that hardly any of them let you customize the number of layers.  I'm not sure why they don't.  It seems like it would be useful.  The only one that does is GraphConvModel, and the structure of that is specialized enough that it isn't obvious what's the right way to turn it into a residual architecture.

This is still a useful feature, just a smaller one than I originally intended.

PR #1700, opened 2019-10-22
Test run when building conda packages was broken

peastman:


peastman:
I'm going to go ahead and merge this without waiting for tests to run, because this code only gets run while building conda packages.

PR #1786, opened 2020-03-26
Be more tolerant of dtypes when computing loss

peastman:


peastman:
All tests are passing except the expected ones.

PR #1875, opened 2020-05-30
fix joblib error for sklearn>0.22

nd-02110114:
I replace `sklearn.externals.joblib` to `joblib` and  move `rdkit` dependency inside the function.


nd-02110114:
I made some mistakes, so I close this PR.

PR #1886, opened 2020-06-06
[WIP] Monorepo

lilleswing:
A possibility of code organization for moving to a single repo.

```
src
├── __pycache__
├── deepchem
│   ├── deepchem # All the code from {git_root}/deepchem
│   └── setup.py
└── tensorchem
    ├── deepchem
    └── setup.py

```

Using pep [420](https://www.python.org/dev/peps/pep-0420/) we can then gradually move all the code that requires tensorflow to tensorchem.  

This has the advantage that they would build different .whl files and thus can be installed independently.

Still WIP playing around with test and seeing if it actually works.  Even if it does it will require massive code movements.


peastman:
> This has the advantage that they would build different .whl files and thus can be installed independently.

Does it matter if they're installed independently?  It's not like it will have much effect on the download size to omit a few packages.  Once we have the import guards set up, you'll be able to just install a single deepchem package.  Any submodules that require a framework you don't have installed just won't get imported.

lilleswing:
@peastman yeah I agree -- I wanted to see what changes would be needed to have multiple packages, and in my opinion they are too drastic.

PR #1915, opened 2020-06-22
[WIP] Adding inorganic crystal datasets

ncfrey:
### Summary
This WIP PR adds experimental and theoretical inorganic crystal datasets to MoleculeNet. Additional featurizers and utility functions for handling crystal structures are added.
- Chemical fingerprint, Sine Coulomb Matrix, and Structure Graph featurizers
- Utilities for loading datasets from json files
- Datasets on metallicity, band gap, and heats of formation

### To-do
- Add datasets to AWS
- Add methods for further featurizing structure graphs

### Additional dependencies
`deepchem.feat.materials_featurizers.py` introduces dependencies on `pymatgen` and `matminer`.


ncfrey:
> This is looking really cool! I'm really excited to see the addition of more materials science support in DeepChem :)
> 
> I did a first pass and have a couple of general comments. We're adding in a lot of new code here so it would be good to break this into a couple PRs. Here's a suggested structure:
> 
> 1. DataLoader PR: It would be useful to factor all the loader changes and tests into one small PR. Looking at the loader code, we might want to think a bit about how we're structuring the loaders. Typically a loader is associated with a filetype, so `JSONLoader` makes sense, but we should think about `CSVCompositionLoader` and `JSONStructureLoader`. These might make sense as their own classes but I'm not sure yet. Factoring out will help us think through it :)
> 2. Materials Featurizers: This would be a clean PR in its own right.
> 3. MoleculeNet chagnes: This would be another clean PR in its own right
> 
> I know this is just WIP but I wanted to give you early feedback so this doesn't turn into the situation I have in #1792 (where I have a gigantic PR that I've been rebasing constantly and painfully trying to factor into smaller PRs). Hope this is useful!

Very useful, thanks for the feedback! I will go ahead and split into three more manageable PRs.

PR #1918, opened 2020-06-23
Remove fit function in graphconv

vsomnath:
This removes the fit function from GraphConv. It directly calls the KerasModel fit, and does not return the loss, causing a None value when running `GraphConvModel.fit()`


peastman:
Looks good.

PR #2019, opened 2020-07-16
[WIP] Split data_loader.py 

nd-02110114:
I just split `data_loader.py` to each file. This PR is a starting point for refactoring Dataloaders like #2009 
And, I think we should remove the load_XXX function in `deepchem/utils/save.py`.


nd-02110114:
I resubmit the PR

PR #2020, opened 2020-07-16
Descriptive material featurizer names

ncfrey:
This PR addresses comments from #2008 with more specific names for inorganic material featurizers, to avoid confusion with other possible types of crystal structures. It also removes `pymatgen` dependent type hints to fix `mypy` errors.


peastman:
Looks good, aside from the mypy errors.  If you intended that users should be able to pass either a single structure or a sequence of multiple structures to `featurize()`, you could specify the type as `OneOrMany[Dict[str, Any]]`.  But looking at the parent class, it looks like the intention is that you always pass in an iterable.  In fact, I think as soon as we add type annotations to the parent class you'll start getting another failure because the argument name isn't consistent.  I think we want the parent class definition to be

```python
def featurize(self, datapoints: Iterable[Any], log_every_n: int = 1000):
```

And then MaterialStructureFeaturizer can override it as

```python
def featurize(self, datapoints: Iterable[Dict[str, Any]], log_every_n: int = 1000):
```

ncfrey:
I added type hints to `Featurizer`, which caused mypy to complain about that class not having `_featurize`, because `featurize` calls `_featurize`, so I added that as well. I fixed the other mypy errors so hopefully it's good to go now.

It seems that mypy doesn't mind if the subclasses change argument names. Currently, `MolecularFeaturizer` uses `molecules` as an argument rather than `datapoints`. I can change everything to `datapoints` though if that would be more clear.

deepchem/feat/base_classes.py
```diff
@@ -289,12 +291,37 @@ def featurize(self, structures: Iterable[JSON],
...
+    """Calculate features for crystal structures.
+
+    Parameters
+    ----------
+    structures: Iterable[Dict[str, Any]]
+      An iterable of pymatgen.Structure dictionaries.
+
+    """
+
+    return self.featurize(structures)
```
nd-02110114, line 316: This override is not necessary. I think the classes in base_classes.py should just override `featurize` basically.

If you want to add these lines and specify arguments  of `__call__` or `_featurize`, this featurizer no longer needs to inherit `Featurizer`.

deepchem/feat/base_classes.py
```diff
@@ -355,6 +381,30 @@ def featurize(self, compositions: Iterable[str],
...
+    """Calculate features for crystal compositions.
+
+    Parameters
+    ----------
+    compositions: Iterable[str]
+      An iterable of crystal compositions.
+
+    """
+
+    return self.featurize(compositions)
```
nd-02110114, line 406: Same here

deepchem/feat/base_classes.py
```diff
@@ -355,6 +368,18 @@ def featurize(self, compositions: Iterable[str],
...
```
nd-02110114, line 381: Same here 🙇

PR #2085, opened 2020-08-11
Fix install scripts for MacOS

nd-02110114:
I fixed the install command of PyTorch for MacOS.

### DIFF
--- a/scripts/install_deepchem_conda.ps1
+++ b/scripts/install_deepchem_conda.ps1
@@ -8,18 +8,7 @@ else
     $python_version=3.6
 }
 
-if($args.Length -eq 1)
-{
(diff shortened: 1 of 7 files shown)


scripts/install_deepchem_conda.ps1
```diff
@@ -8,18 +8,7 @@ else
     $python_version=3.6
 }
 
-if($args.Length -eq 1)
```
nd-02110114, line 11: Currently, we support only `deepchem` virutal  environment because `requirements.yml` support only deepchem env.

See: https://github.com/deepchem/deepchem/blob/master/requirements.yml

nd-02110114:
This is ready to review!

PR #2110, opened 2020-08-21
Update sklearn and xgboost model type annotation

nd-02110114:
## What I did

- Replace `__init__.py` to s`sklearn_model.py` and update type annotation and docstrings
- Replace `__init__.py` to `xgboost_model.py` and update type annotation and docstrings


nd-02110114:
If the CI pass, this PR is ready to merge.

PR #2113, opened 2020-08-22
[WIP] Smiles Tokenizer in dc.feat

seyonechithrananda:
Added new tokenizer for SMILES based off the RXNFP tokenizer. See #2076 for more details

The tokenizer loads its vocab from the vocab.txt found in 'deepchem/feat/tests/data/', is there a better place to place this?

Will also add the import guard shortly.

@rbharath @peastman


deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+    default_vocab_path = (
+        pkg_resources.resource_filename(
+            "chemberta",
+            "tokenizers/vocab.txt"
+        )
+    )
+    return SmilesTokenizer(default_vocab_path)
+
+class SmilesTokenizer(BertTokenizer):
+    r"""
```
rbharath, line 27: Could you swap this docstring over to Numpy documentation style? https://numpydoc.readthedocs.io/en/latest/format.html

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+import re
+import numpy as np
+import pkg_resources
+from typing import List
+from transformers import BertTokenizer
+
+# export
+SMI_REGEX_PATTERN = r"(\[[^\]]+]|Br?|Cl?|N|O|S|P|F|I|b|c|n|o|s|p|\(|\)|\.|=|#|-|\+|\\|\/|:|~|@|\?|>>?|\*|\$|\%[0-9]{2}|[0-9])"
+
+def get_default_tokenizer():
```
rbharath, line 17: Could you add a docstring here?

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+class SmilesTokenizer(BertTokenizer):
+    r"""
+    Constructs a SmilesTokenizer.
+    Bulk of code is from https://github.com/huggingface/transformers and https://github.com/rxn4chemistry/rxnfp
+    Args:
+        vocab_file: Path to a SMILES character per line vocabulary file
+    """
+
+    def __init__(
+            self,
```
rbharath, line 35: Could you add type annotations here? Check out https://deepchem.readthedocs.io/en/latest/coding.html#type-annotations

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def _convert_token_to_id(self, token):
+        """ Converts a token (str/unicode) in an id using the vocab. """
+        return self.vocab.get(token, self.vocab.get(self.unk_token))
+
+    def _convert_id_to_token(self, index):
+        """Converts an index (integer) in a token (string/unicode) using the vocab."""
+        return self.ids_to_tokens.get(index, self.unk_token)
+
+    def convert_tokens_to_string(self, tokens):
```
rbharath, line 90: Could you switch to Numpy doc style and add type annotations for these other methods as well?

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+                        "Saving vocabulary to {}: vocabulary indices are not consecutive."
+                        " Please check that the vocabulary is not corrupted!".
+                            format(vocab_file)
+                    )
+                    index = token_index
+                writer.write(token + u"\n")
+                index += 1
+        return (vocab_file,)
+
+class BasicSmilesTokenizer(object):
```
rbharath, line 157: I think we don't need this class right? It seems most folks would just want `SmilesTokenizer`

deepchem/feat/tests/test_smiles_tokenizer.py
```diff
@@ -0,0 +1,21 @@
...
+
+  def test_tokenize(self):
+      current_dir = os.path.dirname(os.path.realpath(__file__))
+      vocab_path = os.path.join(current_dir, 'data', 'vocab.txt')
+
+      model = RobertaForMaskedLM.from_pretrained('seyonec/SMILES_tokenized_PubChem_shard00_50k')
+      model.num_parameters()
+
+      tokenizer = SmilesTokenizer(vocab_path, max_len=model.config.max_position_embeddings)
+      print(tokenizer.encode("CCC(CC)COC(=O)[C@H](C)N[P@](=O)(OC[C@H]1O[C@](C#N)([C@H](O)[C@@H]1O)C1=CC=C2N1N=CN=C2N)OC1=CC=CC=C1"))
```
rbharath, line 20: Could you add some assertions into this test? It would be good to add checks that the tokenizer has ocrrect behavior and have assertions to guard it

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+def get_default_tokenizer():
+    default_vocab_path = (
+        pkg_resources.resource_filename(
+            "chemberta",
+            "tokenizers/vocab.txt"
+        )
+    )
+    return SmilesTokenizer(default_vocab_path)
+
+class SmilesTokenizer(BertTokenizer):
```
rbharath, line 26: Could you add a file to the docs for tokenizers? Perhaps a new file `docs/tokenizers.rst` with description of our tokenizer infrastructure and autodoc for the new `SmilesTokenizer` class

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+    r"""
+    Constructs a SmilesTokenizer.
+    Bulk of code is from https://github.com/huggingface/transformers and https://github.com/rxn4chemistry/rxnfp
+    Args:
```
rbharath, line 30: Could you add a usage example here showing how to invoke the tokenizer?

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: The SmilesTokenizer uses BasicSmilesTokenizer

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: will do, i believe the test has an example use-case we can port over here

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: will do, we can copy over the bulk of the berttokenizer docs here as they share the same functions

seyonechithrananda:
@rbharath thanks for the comments! Will look over them in further detail tomorrow and address them.

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+        pkg_resources.resource_filename(
+            "chemberta",
+            "tokenizers/vocab.txt"
+        )
+    )
+    return SmilesTokenizer(default_vocab_path)
+
+class SmilesTokenizer(BertTokenizer):
+    r"""
+    Constructs a SmilesTokenizer.
```
peastman, line 28: This is the documentation for the class, not the constructor.

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,183 @@
...
+            self,
+            vocab_file='',
+            # unk_token="[UNK]",
+            # sep_token="[SEP]",
+            # pad_token="[PAD]",
+            # cls_token="[CLS]",
+            # mask_token="[MASK]",
+            **kwargs
+    ):
+        """Constructs a BertTokenizer.
```
peastman, line 44: That should be SmilesTokenizer.

peastman:
A lot more documentation would be good.  What is a SmilesTokenizer, what do you use it for, what algorithm does it use, how do you use it, etc.  A lot of the current docstrings are also pretty uninformative.  Like

```python
"""Run basic SMILES tokenization"""
```

or

```python
"""
Adds special tokens to the a sequence for sequence classification tasks.
A BERT sequence has the following format: [CLS] X [SEP]
"""
```

If you already understand what the function does, I assume that makes sense.  If you don't already understand, it leaves you just as confused as before!

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: done!

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: done!

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: done!

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Please see https://github.com/huggingface/transformers
+    and https://github.com/rxn4chemistry/rxnfp for more details.
+
+    References
+    ----------
+    .. [1]  Schwaller, Philippe; Probst, Daniel; Vaucher, Alain C.; Nair, Vishnu H; Kreutter, David;
+            Laino, Teodoro; et al. (2019): Mapping the Space of Chemical Reactions using Attention-Based Neural
+            Networks. ChemRxiv. Preprint. https://doi.org/10.26434/chemrxiv.9897365.v3
+    Note
```
rbharath, line 40: This should be "Notes"  instead of "Note"

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+        out_string = " ".join(tokens).replace(" ##", "").strip()
+        return out_string
+
+    def add_special_tokens_ids_single_sequence(self, token_ids):
+        """
+        Adds special tokens to the a sequence for sequence classification tasks.
+        A BERT sequence has the following format: [CLS] X [SEP]
+
+
```
rbharath, line 145: Could you add "Parameters" sections to this method and the following methods?

deepchem/feat/smiles_tokenizer.py
rbharath: Ah I see! In that case I suggest renaming this class `_BasicSmilesTokenizer` since this is a private class that shouldn't be part of the public developer facing API

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,235 @@
...
+    return SmilesTokenizer(default_vocab_path)
+
+class SmilesTokenizer(BertTokenizer):
+    """
+    Constructs a SmilesTokenizer. The tokenizer heavily inherits from the BERT
+    WordPieceTokenizer implementation found in Huggingface's transformers library.
+
+    Please see https://github.com/huggingface/transformers
+    and https://github.com/rxn4chemistry/rxnfp for more details.
+
```
rbharath, line 34: Could you add a usage "Examples" section here?

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: It has some functionality, namely that it runs a basic regex SMILES tokenization for those uninterested in using a WordPiece-based tokenization strategy + Huggingface's infrastructure (saving tokenizers, vocab path, etc.). I still think its worth to keep public.

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: done!

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: done

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: fixed

seyonechithrananda:
> A lot more documentation would be good. What is a SmilesTokenizer, what do you use it for, what algorithm does it use, how do you use it, etc. A lot of the current docstrings are also pretty uninformative. Like
> 
> ```python
> """Run basic SMILES tokenization"""
> ```
> 
> or
> 
> ```python
> """
> Adds special tokens to the a sequence for sequence classification tasks.
> A BERT sequence has the following format: [CLS] X [SEP]
> """
> ```
> 
> If you already understand what the function does, I assume that makes sense. If you don't already understand, it leaves you just as confused as before!

@peastman Thanks for the feedback! I've added a new docs page called ```tokenizers.rst``` which describes the tokenization algorithm in more detail to address this.

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: done!

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: done

seyonechithrananda:
@rbharath I've finished adding more changes in response to all the comments. Please let me know if you have any thoughts, or if its ready to be merged in!

deepchem/feat/tests/test_smiles_tokenizer.py
rbharath: New assert looks good! Could you remove the print statement now that we have the assert in?

docs/tokenizers.rst
```diff
@@ -0,0 +1,32 @@
...
+
+- Managing special tokens like mask, beginning-of-sentence, etc tokens (adding them, assigning them to attributes in the tokenizer for easy access and making sure they are not split during tokenization)
+
+BatchEncoding holds the output of the tokenizer’s encoding methods (__call__, encode_plus and batch_encode_plus) and is derived from a Python dictionary. When the tokenizer is a pure python tokenizer, this class behave just like a standard python dictionary and hold the various model inputs computed by these methodes (input_ids, attention_mask…).
+
+For more details on the base tokenizers which the DeepChem tokenizers inherit from, please refer to the following: `HuggingFace tokenizers docs <https://huggingface.co/transformers/main_classes/tokenizer.html>`_
+
+
+SmilesTokenizer
+^^^^^^^^^^^
```
rbharath, line 22: Typo: Can you extend this out to:

```
SmilesTokenizer
^^^^^^^^^^^^^^^^^
```

docs/tokenizers.rst
```diff
@@ -0,0 +1,32 @@
+Tokenizers
+===========
+
+A tokenizer is in charge of preparing the inputs for a model. The HuggingFace transformers library (which DeepChem tokenizers are built on top of) comprise tokenizers for all transformer models.
+
+The base classes PreTrainedTokenizer and PreTrainedTokenizerFast implements the common methods for encoding string inputs in model inputs and instantiating/saving python tokenizers either from a local file or directory or from a pretrained tokenizer provided by the library (downloaded from HuggingFace’s AWS S3 repository).
```
rbharath, line 6: Are `PreTrainedTokenizer` and `PreTrainedTokenizerFast` HuggingFace classes? If so, could you give the full module name `transformers.X.Y.PretrainedTokenizer` and a link to their API documentation?

docs/tokenizers.rst
```diff
@@ -0,0 +1,32 @@
+Tokenizers
```
rbharath, line 1: Can you add a link to `tokenizers.rst` from the main `index.rst` file?

docs/tokenizers.rst
```diff
@@ -0,0 +1,32 @@
...
+^^^^^^^^^^^
+
+The :code:`dc.feat.SmilesTokenizer` module inherits from the BertTokenizer class. It runs a WordPiece tokenization algorithm over SMILES strings using the tokenisation SMILES regex developed by Schwaller et. al.
+
+References:
+
+-  `RXN Mapper: Unsupervised Attention-Guided Atom-Mapping <https://chemrxiv.org/articles/Unsupervised_Attention-Guided_Atom-Mapping/12298559>`_
+-  `Molecular Transformer: Unsupervised Attention-Guided Atom-Mapping <https://pubs.acs.org/doi/10.1021/acscentsci.9b00576>`_
+
+.. autoclass:: deepchem.feat.SmilesTokenizer
```
rbharath, line 31: Can you add autodocs for `BasicSmilesTokenizer` as well?

docs/tokenizers.rst
```diff
@@ -0,0 +1,32 @@
+Tokenizers
+===========
+
```
rbharath, line 3: It might be useful here to add an introductory paragraph explaining why tokenizers could be useful for DeepChem models (perhaps linking to the tutorial for ChemBerta) and providing a bit of broader context for why tokenizers could be powerful for other scientific applications like genomics etc.

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+        ----------
+        token_0: str
+        token_1: str
+        """
+
+        sep = [self.sep_token]
+        cls = [self.cls_token]
+        return cls + token_0 + sep + token_1 + sep
+
+    def add_special_tokens_ids_sequence_pair(self, token_ids_0, token_ids_1):
```
rbharath, line 196: Can you add type annotations, return signatures, and a "Returns" section to the docstring here?

docs/tokenizers.rst
rbharath: I'd also suggest building the docs locally on your machine to check that the hyperlinks/rendering all come out cleanly

deepchem/feat/smiles_tokenizer.py
rbharath: I might be missing it, but looks like `get_default_tokenizer()` doesn't have a docstring yet?

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+# The vocab may be expanded in the near future
+
+import collections
+import logging
+import os
+import re
+import numpy as np
+import pkg_resources
+from typing import List
+from transformers import BertTokenizer
```
rbharath, line 12: Can you expose `SmilesTokenizer` and `BasicSmilesTokenizer` in the top level `deepchem/feat/__init__.py` with an import guard?

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+            Networks. ChemRxiv. Preprint. https://doi.org/10.26434/chemrxiv.9897365.v3
+    Notes
+    ----
+    This class requires huggingface's transformers and tokenizers libraries to be installed.
+
+    """
+    def __init__(
+            self,
+            vocab_file: str='',
+            # unk_token="[UNK]",
```
rbharath, line 61: Should we still have these commented out arguments here?

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+    --------
+
+    >>> from deepchem.feat.smiles_tokenizer import SmilesTokenizer
+    >>> from transformers import RobertaForMaskedLM
+
+    >>> current_dir = os.path.dirname(os.path.realpath(__file__))
+    >>> vocab_path = os.path.join(current_dir, 'data', 'vocab.txt')
+
+    >>> tokenizer = SmilesTokenizer(vocab_path)
+    >>> print(tokenizer.encode("CCC(CC)COC(=O)[C@H](C)N[P@](=O)(OC[C@H]1O[C@](C#N)([C@H](O)[C@@H]1O)C1=CC=C2N1N=CN=C2N)OC1=CC=CC=C1"))
```
rbharath, line 45: The doctest might be failing here. Try running this file with 
```
python -m doctest smiles_tokenizer.py
```
to check that the doctest is running correctly (Travis will also complain)

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+            **kwargs
+    ):
+
+        """Constructs a SmilesTokenizer.
+
+        Parameters
+        ----------
+        vocab_file: str
+            Path to a SMILES character per line vocabulary file.
+            Default vocab file is found in deepchem/feat/tests/data/vocab.txt
```
rbharath, line 75: A couple options here. We could make the vocab file into a source file (we define a python variable with the vocabulary). Or perhaps we could move the vocab file up to S3 and download it on need.

CC @peastman who might have some suggestions on this

docs/tokenizers.rst
seyonechithrananda: Yup, pycharm shows the individual docs, but I can check out how It looks on ReadtheDocs too

seyonechithrananda:
Thanks, @rbharath for the 2nd pass! I will go through your comments and address them!

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: woops confused docstring with pep8 code styling lol

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+    WordPieceTokenizer implementation found in Huggingface's transformers library.
+
+    Please see https://github.com/huggingface/transformers
+    and https://github.com/rxn4chemistry/rxnfp for more details.
+
+    Examples
+    --------
+
+    >>> from deepchem.feat.smiles_tokenizer import SmilesTokenizer
+    >>> from transformers import RobertaForMaskedLM
```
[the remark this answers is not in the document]
seyonechithrananda: done

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+
+    @property
+    def vocab_size(self):
+        return len(self.vocab)
+
+    @property
+    def vocab_list(self):
+        return list(self.vocab.keys())
+
+    def _tokenize(self, text):
```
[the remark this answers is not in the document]
seyonechithrananda: done

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+        """
+        Tokenize a string into a list of tokens.
+
+        Parameters
+        ----------
+        text: str
```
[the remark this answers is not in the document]
seyonechithrananda: done

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,292 @@
...
+        right: bool (True by default)
+
+        """
+        padding = [self.pad_token_id] * (length - len(token_ids))
+        if right:
+            return token_ids + padding
+        else:
+            return padding + token_ids
+
+    def save_vocabulary(self, vocab_path):
```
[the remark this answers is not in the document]
seyonechithrananda: done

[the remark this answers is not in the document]
seyonechithrananda: return type here is Tuple[str]

docs/tokenizers.rst
seyonechithrananda: done

docs/tokenizers.rst
seyonechithrananda: yup they are, i linked the API documentation below in the docs

docs/tokenizers.rst
seyonechithrananda: done

docs/tokenizers.rst
seyonechithrananda: done

docs/tokenizers.rst
seyonechithrananda: done

docs/tokenizers.rst
seyonechithrananda: done

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: done

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: for this, would the following blurb work?

```
try:
    import transformers
    from transformers import BertTokenizer

    from deepchem.feat.smiles_tokenizer import SmilesTokenizer
    from deepchem.feat.smiles_tokenizer import BasicSmilesTokenizer
except ImportError:
    logger.warning("HuggingFace transformers is not available. Please install using 'pip install transformers' to use the SmilesTokenizer")
```

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: might be good to keep to preserve how the original wordpiece tokenizer works!

seyonechithrananda: can remove if needed

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: Sounds good, might want to make it a text file for people to share if needed

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: added this for now, feel free to let me know if any changes necessary

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: Ran this and fixed all errors.

deepchem/feat/__init__.py
```diff
@@ -25,3 +25,13 @@
 from deepchem.feat.material_featurizers import ElementPropertyFingerprint
 from deepchem.feat.material_featurizers import SineCoulombMatrix
 from deepchem.feat.material_featurizers import CGCNNFeaturizer
+
+try:
+    import transformers
```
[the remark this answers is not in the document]
rbharath: I think running `yapf` might fix the Travis CI failure we're seeing here as well (https://travis-ci.org/github/deepchem/deepchem/jobs/721857444#L2686)

deepchem/feat/__init__.py
```diff
@@ -25,3 +25,13 @@
...
+    from transformers import BertTokenizer
+
+    from deepchem.feat.smiles_tokenizer import SmilesTokenizer
+    from deepchem.feat.smiles_tokenizer import BasicSmilesTokenizer
+except ImportError:
```
rbharath, line 35: I think you might need `ModuleNotFoundError` here instead of `ImportError`. See this example import guard: https://github.com/deepchem/deepchem/blob/master/deepchem/models/__init__.py#L40

deepchem/feat/smiles_tokenizer.py
rbharath: Ah, so it's documentation about the vocab file format! Makes sense. Can you move this into into the docstring below so the vocab file format shows up in the rendered docs?

deepchem/feat/smiles_tokenizer.py
rbharath: Type annotation still missing here and on some of the other methods

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,357 @@
...
+
+        token_ids_1: List[int]
+            List of tokens for the second string sequence in the sequence pair (B).
+        """
+
+        sep = [self.sep_token_id]
+        cls = [self.cls_token_id]
+        return cls + token_ids_0 + sep + token_ids_1 + sep
+
+    def add_padding_tokens(self, token_ids, length, right=True):
```
rbharath, line 242: Type annotation

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,357 @@
...
+    Examples
+    --------
+    >>> from deepchem.feat.smiles_tokenizer import BasicSmilesTokenizer
+
+    >>> tokenizer = BasicSmilesTokenizer()
+    >>> print(tokenizer.tokenize("CCC(CC)COC(=O)[C@H](C)N[P@](=O)(OC[C@H]1O[C@](C#N)([C@H](O)[C@@H]1O)C1=CC=C2N1N=CN=C2N)OC1=CC=CC=C1"))
+    ['C', 'C', 'C', '(', 'C', 'C', ')', 'C', 'O', 'C', '(', '=', 'O', ')', '[C@H]', '(', 'C', ')', 'N', '[P@]', '(', '=', 'O', ')', '(', 'O', 'C', '[C@H]', '1', 'O', '[C@]', '(', 'C', 'N', ')', '(', '[C@H]', '(', 'O', ')', '[C@@H]', '1', 'O', ')', 'C', '1', '=', 'C', 'C', '=', 'C', '2', 'N', '1', 'N', '=', 'C', 'N', '=', 'C', '2', 'N', ')', 'O', 'C', '1', '=', 'C', 'C', '=', 'C', 'C', '=', 'C', '1']
+
+
+    References
```
rbharath, line 322: Formatting. This should be:

```
References
----------
```

docs/tokenizers.rst
rbharath: Perhaps we could clarify a bit here. Something like 

"A tokenizer is in charge of preparing the inputs for a natural language processing model. For many scientific applications, it is possible to treat inputs as "words"/"sentences" and use NLP methods to make meaningful predictions. For example, SMILES strings or DNA sequences have grammatical structure and can be usefully modeled with NLP techniques. DeepChem provides some scientifically relevant tokenizers for use in different applications. These tokenizers are based on those from the Huggingface transformers library (which DeepChem tokenizers inherit from)."

deepchem/feat/smiles_tokenizer.py
rbharath: Looks like the docstring for `get_default_tokenizer()` still needs to be added here

deepchem/feat/smiles_tokenizer.py
rbharath: See my comment above (basically, we should run `yapf` and use `ModuleNotFoundError` instead of `ImportError`

rbharath:
Forgot to mention in my review, we'll also have to install transformers so that Travis can run the transformers test. You can do that by modifying `scripts/install_deepchem_conda.sh` and `scripts/install_deepchem_conda.ps1` to add a line for `pip install transformers`.  For example, see how pytorch is installed (https://github.com/deepchem/deepchem/blob/master/scripts/install_deepchem_conda.sh#L59, https://github.com/deepchem/deepchem/blob/master/scripts/install_deepchem_conda.ps1#L50)

deepchem/feat/__init__.py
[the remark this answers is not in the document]
seyonechithrananda: done

deepchem/feat/__init__.py
seyonechithrananda: done

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: added return type annotation to add_special_tokens_ids_sequence_pair()

seyonechithrananda:
Thanks, @rbharath for the really helpful feedback here! Implemented all the changes, there are no longer any methods without type annotations, the code is styled with yapf, and I added transformers to the sh + ps1 scripts for Travis CI!

Thumbs crossed we should be good to go haha, let's hope Travis doesn't complain again

nd-02110114:
Before merging in this PR, I want to confirm why this implementation is needed in DeepChem. 

Before reviewing, I confirmed the original implementation. (See https://github.com/rxn4chemistry/rxnfp/blob/master/rxnfp/tokenization.py)

Your codes is completely same as `rxnfp/tokenization.py`. It is not necessary to migrate these codes to DeepChem, just we should install rxnfp packages via pypi or conda and use their functions or classes like SmilesTokenizer, BasicSmilesTokenizer.  In addition to this, I think the migration of `rxnfp/tokenization.py` has a risk that it is difficult to keep up with the latest bugfix in rxnfp repository.

seyonechithrananda:
@nd-02110114 Thanks for the comments! I've been working on a ChemBERTa implementation, and so I thought it would be good to have the tokenizer integrated into the library such that we can use components in the library for ChemBERTa + the tutorial notebook.  (see #2050)

I spoke to Phillipe (the co-author of the rxnfp repo) and it seems like that will be the stable release of RXNFP, which is unlikely to see an update, as it is just a default tokenizer from an older research work (RXNMapper: https://github.com/rxn4chemistry/rxnmapper). It's unlikely to be changed for this reason.

nd-02110114:
@seyonechithrananda 
> so I thought it would be good to have the tokenizer integrated into the library such that we can use components in the library for ChemBERTa once it is released in DeepChem using TorchModel.

Sorry, I couldn't understand benefits about integrating some rxnfp functions or classes into DeepChem because we can install and import rxnfp easily using conda or pypi. And, DeepChem basically serves many useful Featurizers by wrapping classes or functions of external packages like Pymatgen, RDKit and so on, not directly migrating some codes to DeepChem. (See other featurizer like CirCularFingerprint or ElementPropertyFingerprint and so on). So, I think that it is better to create the original featurizer by importing rxnfp and using some functions or classes than to migrate all codes to DeepChem if the implementation is completely same as the `rxnfp/tokenization.py`. The advantage of this style is to reduce the codes which we should maintenance. 

If you don't  (or can't ) create the orignal featurizer for some reasons but you really want to put some rxnfp functions or classes on DeepChem, I think it is better to just export them from `deepchem.feat` like the below.

https://github.com/deepchem/deepchem/blob/bc05e7c116327da590bb1b64a1a63dba7050ee17/deepchem/metrics/__init__.py#L14-L30

DeepChem export some useful scikit-learn or scipy functions without no modification.
 
> I spoke to Phillipe (the co-author of the rxnfp repo) and it seems like that will be the stable release of RXNFP, which is unlikely to see an update, as it is just a default tokenizer from an older research work (RXNMapper: https://github.com/rxn4chemistry/rxnmapper). It's unlikely to be changed for this reason.

I could understand the situation of rxnfp. 

And , If you want to add `SmilesTokenizer` to DeepChem, I think it should be available with dataloader. The current design of `SmilesTokenizer` doesn't inherit any base class of Featurizer. I think `SmilesTokenizer` should inherit `MolecularFeaturizer` and the `tokenize` method of `SmilesTokenizer` is converted to `_featurize` method.

seyonechithrananda:
Sounds good! I'll take a look if we can inherit from MolecularFeaturizer. Right now, the SmilesTokenizer overwrites the majority of the BertTokenizer methods and just calls the BasicSmilesTokenizer, which runs a regex tokenization pattern. Because of this its fairly robust and unlikely to be of high maintenance (because even if the transformers library changes, the tokenizer uses regex, not transformers' WordPiece tokenization algorithm). Will also talk with @rbharath when he is back next week and get his thoughts.

seyonechithrananda:
The travis CI is still failing because of an error with the logger, will fix now.

seyonechithrananda:
The reason the CI still fails seems to be some issue with type annotations.  It expects `token_0` to be a List in line 226 of `smiles_tokenizer.py` although I explicitly define it as an`str`: 

Screen Shot 2020-08-30 at 5 28 04 PM
Screen Shot 2020-08-30 at 5 28 19 PM

@rbharath Any thoughts as to how to resolve this? Other than this, it should pass and be ready for merging.

peastman:
I think we still need more documentation.  As an example, the docstring for the SmilesTokenizer class just says,

> Creates the SmilesTokenizer class. The tokenizer heavily inherits from the BERT
> WordPieceTokenizer implementation found in Huggingface's transformers library.

That doesn't answer the basic questions a user will have when reading through the documentation: what is this class for, is it useful for solving the problem I'm working on right now, etc.  It does include links to documentation on other sites, but users shouldn't have to follow links just to find out what a class does.  The description of it in `tokenizers.rst` is only slightly more userful:

> The :code:`dc.feat.SmilesTokenizer` module inherits from the BertTokenizer class. It runs a WordPiece tokenization
> algorithm over SMILES strings using the tokenisation SMILES regex developed by Schwaller et. al.

Linking to the original article for advanced details is good, but we should still give a basic description in the manual.

seyonechithrananda:
> I think we still need more documentation. As an example, the docstring for the SmilesTokenizer class just says,
> 
> > Creates the SmilesTokenizer class. The tokenizer heavily inherits from the BERT
> > WordPieceTokenizer implementation found in Huggingface's transformers library.
> 
> That doesn't answer the basic questions a user will have when reading through the documentation: what is this class for, is it useful for solving the problem I'm working on right now, etc. It does include links to documentation on other sites, but users shouldn't have to follow links just to find out what a class does. The description of it in `tokenizers.rst` is only slightly more userful:
> 
> > The :code:`dc.feat.SmilesTokenizer` module inherits from the BertTokenizer class. It runs a WordPiece tokenization
> > algorithm over SMILES strings using the tokenisation SMILES regex developed by Schwaller et. al.
> 
> Linking to the original article for advanced details is good, but we should still give a basic description in the manual.

Will do! I also found a bug with the save_vocablary method, fixing that right now. Once we update that I'll update the docs

deepchem/feat/smiles_tokenizer.py
```diff
@@ -0,0 +1,369 @@
...
+from logging import getLogger
+
+logger = getLogger(__name__)
+
+
+try:
+  from transformers import BertTokenizer
+except ModuleNotFoundError:
+  logger.warning("HuggingFace transformers is not available. Please install using 'pip install transformers' to use the SmilesTokenizer")
+  print("HuggingFace transformers is not available. Please install using 'pip install transformers' to use the SmilesTokenizer.")
```
rbharath, line 23: Could you remove the print statement here? (For production use cases, we should make sure we only use `logger` to output statements). Also, no need for the `pass statement`

peastman:
Conceptually, tokenizers are a type of featurizer.  That suggests to me we need *some* way to use them as featurizers, and to plug them into all the places where featurizers are allowed.  Having the tokenizer subclass MolecularFeaturizer would be one way of doing it.  An alternative would be to do it through encapsulation, for example with a TokenizingFeaturizer class that takes a tokenizer as an argument and uses it to perform featurization.  I can see advantages to both approaches.

deepchem/feat/smiles_tokenizer.py
seyonechithrananda: will do

seyonechithrananda: done

seyonechithrananda:
@rbharath @peastman 

We inherit the `BertTokenizer` from HuggingFace's transformers library, which inherits from the `PreTrainedTokenizer`. You can tokenize entire datasets by feeding a list of sequences and using a for loop to process each string using the `tokenize` method. However, for ChemBERTa we utilize the HuggingFace trainer class for pre-training as well as property prediction tasks, of which you can just pass the tokenizer to the `LinebyLineTextDataset`:

```
from transformers import LineByLineTextDataset

dataset = LineByLineTextDataset(
    tokenizer=tokenizer,
    file_path="./250k_rndm_zinc_drugs_clean_sorted copy.txt",
    block_size=128,
)
```

From here we just pass the dataset to a trainer object, and run `trainer.train()`:
```

trainer = Trainer(
    model=model,
    args=training_args,
    data_collator=data_collator,
    train_dataset=dataset,
    prediction_loss_only=True,
)
```

I think we could create methods for tokenizing deepchem `Dataset` objects, for example, but I think for now HuggingFace has the suitable methods for tokenizing large datasets.

Another example of featurizing large datasets:
```

dataset_path = "/tox21/tox21_balanced_revised_no_id.csv"
df = pd.read_csv(dataset_path, sep = ',', warn_bad_lines=True, header=None)
df.rename(columns={0:'sentence',1:'label'}, inplace=True)
# Get the lists of sentences and their labels.
sentences = df.sentence.values
labels = df.label.values


train_texts, val_texts, test_texts, train_labels, val_labels, test_labels = train_val_test_split(sentences, labels, test_size=.2)

tokenizer = SmilesTokenizer(vocab_path, max_len=model.config.max_position_embeddings)

train_encodings = tokenizer(train_texts, truncation=True, padding=True)
val_encodings = tokenizer(val_texts, truncation=True, padding=True)
test_encodings = tokenizer(test_texts, truncation=True, padding=True)


class Tox21Dataset(torch.utils.data.Dataset):
    def __init__(self, encodings, labels):
        self.encodings = encodings
        self.labels = labels

    def __getitem__(self, idx):
        item = {key: torch.tensor(val[idx]) for key, val in self.encodings.items()}
        item['labels'] = torch.tensor(self.labels[idx])
        return item

    def __len__(self):
        return len(self.labels)

train_dataset = Tox21Dataset(train_encodings, train_labels)
val_dataset = Tox21Dataset(val_encodings, val_labels)
test_dataset = Tox21Dataset(test_encodings, test_labels)
```

seyonechithrananda:
I've also fixed the `save_vocablary()` method which previously threw errors when saving the `vocab.txt` file. Did another run-through of the unit test and it looks good.

deepchem/feat/__init__.py
```diff
@@ -25,3 +25,17 @@
...
+  from logging import getLogger
+  logger = getLogger(__name__)
+  import transformers
+  from transformers import BertTokenizer
+
+  from deepchem.feat.smiles_tokenizer import SmilesTokenizer
+  from deepchem.feat.smiles_tokenizer import BasicSmilesTokenizer
+except ModuleNotFoundError:
+  logger.warning("HuggingFace transformers is not available. Please install using 'pip install transformers' to use the SmilesTokenizer")
+
```
seyonechithrananda, line 39: @rbharath I removed print statements and opted to use logger here, also removed the pass command.

nd-02110114:
@rbharath Thank you for your summary! I really appreciate 🙇‍♂️ 

> If I'm understanding @seyonechithrananda's point, the main reason to migrate the code into DeepChem is that rxnfp isn't actively being developed further, so it's unlikely to see future improvements. In this case, since we anticipate using tokenizers more heavily moving forward, I think it would make sense to add a DeepChem implementation based off rxnfp's code.

If there is a high possibility to customize the original rxnfp's SmilesTokenizer in the future, I can agree.

> It makes sense to me that we should defer to HuggingFace's API since they're the established leader in this area. But we should have a clear API for tokenizing large datasets. If we decide to follow HuggingFace's lead, we don't need to make tokenizers inherit from MolecularFeaturizer, but we should add a section to tokenizers.rst explaining how to tokenize a large dataset HuggingFace-style and load it into DeepChem.

In my image, SmilesTokenizer should inherits both MolecularFeaturizer and BertTokenizer. 


```python

class SmilesTokenizer(BertTokenizer, MolecularFeaturizer):
  def __init__(
      self,
      vocab_file: str='',
      **kwargs):
    ....

 def _featurize(self, mol: RDKitMol) -> np.ndarray:
    from rdkit import Chem
    smiles = Chem.MolToSmiles(mol)
    return np.array(self.encode(smiles))

 def .....

```

rbharath:
@nd-02110114 I really like this design! It's a light weight addition that would let us use `SmilesTokenizer` in both HuggingFace's and DeepChem's pipelines. @seyonechithrananda What do you think of this API design? I believe it would just require adding the single extra `_featurize` method that @nd-02110114 has implemented above and adding a test case for invoking the `_featurize()` method. I'd be OK doing this in a follow-on PR if we're agreed about the design.

@seyonechithrananda It looks like `yapf` needs to be run and it looks like `test_tokenize` is failing in the test suite. Would you be able to fix these failures? No more code change requests on my end for this PR :)

seyonechithrananda:
@rbharath I agree, I really like @nd-02110114's suggested design. I'll definitely make a follow-up PR with this design once I have some more time, probably around mid-September (after the Arxiv release). That way, we can use more of the DeepChem pipeline (as currently, the SmilesTokenizer works well with Huggingface but not as well with deepchem). Will run yapf again on the unit test + tokenizer script, and look at why it's failing. Should be good to be merged after!

seyonechithrananda:
@rbharath The reason why the CI is failing is because of this type annotation issue: 
image

Is there any way I can resolve it? It seems to expect a list.

seyonechithrananda:
Fixed the CI issue, it's with a specific method, but I believe we can remove it as it corresponds to WordPiece tokenization which we don't use.

I also added YAPF formatting to the tokenizer + unit test and tested the unit test and the assertion passes. Fingers crossed it passes now 🤞

seyonechithrananda:
@rbharath I'm still getting an error with yapf formatting over the __init__.py, but I ran a quick check and its already been formatted properly. Any thoughts on how to fix this? otherwise, the CI is passing. 
Screen Shot 2020-09-02 at 9 00 22 PM

peastman:
Make sure you're using yapf 0.22.  If you have a different version it will format things slightly differently and the test won't pass.

seyonechithrananda:
@peastman Will check! thanks :)

seyonechithrananda:
@peastman I just verified that my version of yapf is indeed 0.22, no changes were suggested over the files though. Any ideas why this is still failing?

rbharath:
@seyonechithrananda Can you try running this sequence of yapf commands from the command line? The sequence I usually use is just

```
yapf --version # to check version
yapf -i /path/to/file
git add /path/to/file
git commit -m "My message"
git push origin branch # Use your remote/branch here
```

I'm not sure what's happening here but perhaps there's something in your system setup that is a little different here so might be good to try the simplest setup to see if that helps

seyonechithrananda:
> @seyonechithrananda Can you try running this sequence of yapf commands from the command line? The sequence I usually use is just
> 
> ```
> yapf --version # to check version
> yapf -i /path/to/file
> git add /path/to/file
> git commit -m "My message"
> git push origin branch # Use your remote/branch here
> ```
> 
> I'm not sure what's happening here but perhaps there's something in your system setup that is a little different here so might be good to try the simplest setup to see if that helps

@rbharath 
version: 0.11.1 (I installed 0.22 using `pip install yapf==0.22`)
When I run `yapf -i` on `smiles_tokenizer.py`, `__init__.py`, and `test_smiles_tokenizer.py`, no changes are reported.

rbharath:
Hmm, that looks like the wrong yapf version. Here's what I see when I run the version command:

```
(deepchem) bharath@Bharaths-MBP tests % yapf --version
yapf 0.22.0
```

seyonechithrananda:
@rbharath Yeah, kinda weird because I did a fresh install of yapf before running again. Will try with a new environment. Did you use the pip command or clone yapf?

rbharath:
I used pip, but might be worth running `which yapf` to make sure that you're using the pip installed yapf in your environment

seyonechithrananda:
@rbharath Just fixed it with the fresh install! Adding changes now

seyonechithrananda:
Lets hope it finally passes now haha, thanks so much for the help @rbharath

rbharath:
Travis is now green so going to merge this in! Congrats on the new feature merged in @seyonechithrananda :)

We should do further work to integrate Tokenizers more closely into DeepChem as indicated in the discussion above in future PRs

PR #2123, opened 2020-08-25
Add mode args for CGCNNModel

nd-02110114:
## What I did

- Add mode option for CGCNNModel like other graph conv models
- Reupload `mp_is_metal.tar.gz` in S3
  - The task value of `mp_is_metal.tar.gz` is bool, so I converted bool value to int (on-hot)
- Update docstrings and fix flake8 error in materials dataset molnet functions
- Small refactor for wandb flag


nd-02110114:
I just added docstrig modifications and improve the doctest logs. So, I merge in.

PR #2138, opened 2020-09-03
Fix GAT and CGCNN support for GPU

nd-02110114:
## What I did

- Fix SparseSoftmaxCrossEntropy loss
  - There is a bug in the case that output is `(batch_size, tasks, classes)` 
- Add mode option and more tests for GATModel
- Fix GATModel for GPU
- Refactor CGCNNModel


nd-02110114:
I refactored GATModel and checked I refactored GATModel and checked whether to work well for classification task and regression task.  

See the notebook https://colab.research.google.com/drive/1eF7WtiLbdTBafP7v1ge4Mvq6F3Y7zmIQ?usp=sharing

I seem that GATModel is not bad although the convergence is slow (a little bit strange). Tox21 mean roc-auc score is around 0.75.

On the other hand, I also refactored CGCNNModel. But, the CGCNNModel still doesn't work well.  The issue is that the loss value doesn't decrease after 1 epochs like this.

```
1.2885267893473307
0.5434092839558919
0.5397125244140625
0.5522617340087891
0.5471174240112304
0.5393458684285481
0.540155553817749
0.5367419560750325
0.5460977554321289
0.566059430440267
```

nd-02110114:
@rbharath 
I continue to debugging the CGCNNModel, but I think this PR is ready to merge. I will make the another PR for more fixes about CGCNNModel

rbharath (approved):
Looks great! Thanks for investigating this @nd-02110114 :).

PR #2147, opened 2020-09-04
[WIP] Remove dataset directory

### LINKED ISSUE #1892: datasets and website directories are necessary...?
I seem `datasets` and `website` directories are not used.... 
Can we remove these? Especially, it is better to remove datasets (about 150Mb).
deepchem repository size is a little big and it takes time to clone.

To be exact, `datasets/mini_muv.csv.gz` (20kb) are only used in test, so we can move `deepchem/data/tests/` like `mini_emol.csv`.

nd-02110114:
Fix #1892 

I remove dataset directory and outdate example files.


rbharath (commented):
We're removing a lot of examples here that have some good content. I'd suggest that we move the simple examples into the docs (and tested under doctest) rather than deleting them outright.

For the datasets, could you verify that all these datasets are in the S3 bucket? If so, perhaps this could be a simple PR that just deletes the dataset directory and doesn't change the examples directory

nd-02110114:
> We're removing a lot of examples here that have some good content. I'd suggest that we move the simple examples into the docs (and tested under doctest) rather than deleting them outright.

You are defintely right! I will go back examples.

> For the datasets, could you verify that all these datasets are in the S3 bucket? 

Yes.

PR #2152, opened 2020-09-08
Fix commented code in tutorial 4

neel-shah:


rbharath (commented):
This looks great! Still a slight dropoff in accuracy for the custom model compared to the wrapped DeepChem model but it's pretty comparable. Will leave this open for a while in case other folks have comments but good to merge on my end

PR #2166, opened 2020-09-14
Deprecating energy features for Transformers 

alat-rights:
- [ X ] Added deprecation warning to untransform_grad method of NormalizationTransformer class. Related to issue #1978

Need to look for other remaining features that should also be deprecated. 

Related to issue #1978 .


rbharath (approved):
Looks like this PR is good to merge! Once you remove the draft status, we should be good to go :)

PR #2167, opened 2020-09-14
QM9 loader bugfix

ncfrey:
While working on #2161, I found a bug in the QM9 loader. I also added some tests.


rbharath (approved):
This is great! Thanks for making the fix :)

PR #2199, opened 2020-10-02
Fix deprecation warnings occuring due to invalid escape sequences

rfhari:
Fixes issue #2007


rfhari:
There were more warnings but not deprecation warnings so I didn't include those changes here. Should I include them as well?
Here is the screenshot

> 
image

rbharath:
It looks like the other errors are from the `contrib/` folder which isn't actively maintained, so this PR is probably good to merge as is. Congrats on your first accepted DeepChem PR :)

rfhari:
Thanks a lot, Bharath!!

PR #2200, opened 2020-10-02
Changes to ButinaSplitter

### LINKED ISSUE #2198: Changes to ButinaSplitter
cc @rbharath @proteneer 

This is a followup to https://github.com/deepchem/deepchem/issues/2112#issuecomment-700942341 and the discussion following it.  ButinaSplitter is currently not very useful, but could be made more useful with a few changes.

As discussed there I modified it to use the same code for building datasets as ScaffoldSplitter.  I then tested it with the following code.

```python
splitters = ['random', 'scaffold', 'butina']
metric = dc.metrics.Metric(dc.metrics.roc_auc_score)
for splitter in splitters:
    tasks, datasets, transformers = dc.molnet.load_tox21(featurizer='ECFP', split=splitter)
    model = dc.models.MultitaskClassifier(n_tasks=len(tasks), n_features=1024, layer_sizes=[1000])
    model.fit(train_dataset, nb_epoch=10)
    print('splitter:', splitter)
    print('training set score:', model.evaluate(train_dataset, [metric], transformers))
    print('test set score:', model.evaluate(test_dataset, [metric], transformers))
    print()
```

RandomSplitter gives a test set score of 0.78, and ScaffoldSplitter gives 0.68, which shows how random splitting doesn't produce a truly independent test set.  Scaffold splitting does much better.  ButinaSplitter gives a test set score of 0.78, exactly the same as RandomSplitter!

ButinaSplitter has an adjustable cutoff.  The above was with the default value of 0.18.  So I repeated the test with different values of the cutoff.  Here are the test set scores.

0.18: 0.78
0.3: 0.74
0.4: 0.73
0.5: 0.66
0.6: 0.61
0.7: 0.65
0.8: 0.61

At least on this dataset, 0.18 is much too small.  It works no better than random splitting.  It gets better as you increase the cutoff, and by 0.6 it's significantly better than ScaffoldSplitter.  Further increases make little difference.

I looked up the original paper on the clustering method (http://www.l4patterns.com/uploads/dbclus-paper.pdf), and they used a cutoff of 0.8.  How was the value 0.18 selected?  Would it be reasonable to increase the default to 0.6?  Is this value likely to be universal, or will it vary a lot between datasets?

peastman:
Fixes #2198.  I made the following changes.

1. It now performs the splitting using identical code to ScaffoldSplitter, just based on clusters instead of scaffolds.
2. I increased the default cutoff to 0.6.
3. The cutoff is now a constructor argument.  This allows `split()` to have an identical signature to other splitters.  It also will allow you to control the cutoff for molnet loader function with, for example, `splitter=dc.split.ButinaSplitter(0.7)`.
4. I updated the splitters tutorial to discuss it as well.

### DIFF
--- a/deepchem/splits/splitters.py
+++ b/deepchem/splits/splitters.py
@@ -1015,14 +1015,26 @@ class ButinaSplitter(Splitter):
   This class requires RDKit to be installed.
   """
 
-  def split(self,
-            dataset: Dataset,
-            frac_train: float = 0.8,
-            frac_valid: float = 0.1,
-            frac_test: float = 0.1,
-            seed: Optional[int] = None,
-            log_every_n: Optional[int] = None,
-            cutoff: float = 0.18) -> Tuple[List[int], List[int], List]:
+  def __init__(self, cutoff: float = 0.6):
+    """Create a ButinaSplitter.
+
+    Parameters
+    ----------
+    cutoff: float (default 0.6)
+      The cutoff value for tanimoto similarity.  Molecules that are more similar
+      than this will tend to be put in the same dataset.
+    """
+    super(ButinaSplitter, self).__init__()
+    self.cutoff = cutoff
+
+  def split(
+      self,
+      dataset: Dataset,
+      frac_train: float = 0.8,
+      frac_valid: float = 0.1,
+      frac_test: float = 0.1,
+      seed: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


rbharath (approved):
Looks good! Feel free to merge when ready :)

PR #2216, opened 2020-10-14
Fix __str__ method for Featurizer

nd-02110114:
I got the error from the following code.

```python
>>> str(dc.feat.Mol2VecFingerprint())
```

So, I fixed this error.

### DIFF
--- a/deepchem/feat/base_classes.py
+++ b/deepchem/feat/base_classes.py
@@ -134,18 +134,19 @@ class Featurizer(object):
 
     override_args_info = ''
     for arg_name, default in zip(args_names, args_default_values):
-      arg_value = self.__dict__[arg_name]
-      # validation
-      # skip list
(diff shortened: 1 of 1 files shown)


deepchem/feat/base_classes.py
```diff
@@ -134,18 +134,19 @@ def __str__(self) -> str:
...
-      arg_value = self.__dict__[arg_name]
-      # validation
-      # skip list
-      if isinstance(arg_value, list):
-        continue
-      if isinstance(arg_value, str):
-        # skip path string
-        if "\\/." in arg_value or "/" in arg_value or '.' in arg_value:
+      if arg_name in self.__dict__:
+        arg_value = self.__dict__[arg_name]
```
nd-02110114, line 138: This line raised the KeyError.

rbharath (approved):
Thanks for the fix!

PR #2227, opened 2020-10-16
Fix inconsistent API for HyperparamOpt

nd-02110114:
I found the inconsistent API between GaussianProcessHyperparamOpt and GridHyperparamOpt. 

`GridHyperparamOpt` have the `output_transformers` argument, but `GaussianProcessHyperparamOpt` doesn't have. I found `GridHyperparamOpt` apply transformers to predicted values, but `GaussianProcessHyperparamOpt` doesn't apply. So, I fixed the `hyperparam_search` method that `GaussianProcessHyperparamOpt` also apply transformers to predicted values.

**GridHyperparamOpt**
https://github.com/deepchem/deepchem/blob/51a76d96dc82aafb22f4b2265d5283763069654a/deepchem/hyper/grid_search.py#L65-L73

**GaussianProcessHyperparamOpt**
https://github.com/deepchem/deepchem/blob/51a76d96dc82aafb22f4b2265d5283763069654a/deepchem/hyper/gaussian_process.py#L131-L141

This modification needs breaking API, so if you have some idea to avoid breaking API, please tell me 🙇

### DIFF
--- a/deepchem/feat/base_classes.py
+++ b/deepchem/feat/base_classes.py
@@ -315,7 +315,7 @@ class MaterialStructureFeaturizer(Featurizer):
   """
 
   def featurize(self,
-                structures: Iterable[Union[Dict, PymatgenStructure]],
+                structures: Iterable[Union[Dict[str, Any], PymatgenStructure]],
                 log_every_n: int = 1000) -> np.ndarray:
     """Calculate features for crystal structures.
 
--- a/deepchem/feat/base_classes.py
+++ b/deepchem/feat/base_classes.py
@@ -233,7 +233,7 @@ class MolecularFeaturizer(Featurizer):
   The subclasses of this class require RDKit to be installed.
   """
 
-  def featurize(self, molecules, log_every_n=1000):
+  def featurize(self, molecules, log_every_n=1000) -> np.ndarray:
     """Calculate features for molecules.
 
     Parameters
@@ -315,7 +315,7 @@ class MaterialStructureFeaturizer(Featurizer):
   """
 
   def featurize(self,
-                structures: Iterable[Union[Dict[str, Any], PymatgenStructure]],
+                structures: Iterable[Union[Dict, PymatgenStructure]],
                 log_every_n: int = 1000) -> np.ndarray:
     """Calculate features for crystal structures.
 
--- a/deepchem/hyper/gaussian_process.py
+++ b/deepchem/hyper/gaussian_process.py
@@ -9,7 +9,6 @@ from typing import Dict, List, Optional, Tuple, Union
 from deepchem.data import Dataset
 from deepchem.trans import Transformer
 from deepchem.metrics import Metric
-from deepchem.utils.evaluate import Evaluator
 from deepchem.hyper.base_classes import HyperparamOpt
 from deepchem.hyper.base_classes import _convert_hyperparam_dict_to_filename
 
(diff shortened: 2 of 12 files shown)


rbharath (commented):
I think this is a nice change! Unifies the API and lets us get rid of those pesky API mismatches across the hyperparameter optimizers. 

I think this change, although it minorly changes signatures, is worth making since it closes a oversight and makes APIs more uniform. 

@peastman Would you have bandwidth to do a quick pass for a second opinion to confirm?

deepchem/hyper/gaussian_process.py
```diff
@@ -280,7 +285,8 @@ def optimizing_function(**placeholders):
       except NotImplementedError:
         pass
 
-      multitask_scores = model.evaluate(valid_dataset, [metric])
+      evaluator = Evaluator(model, valid_dataset, output_transformers)
+      multitask_scores = evaluator.compute_model_performance([metric])
```
peastman, line 289: Why this change?  You could just pass the transformers to `evaluate()` and avoid having to create an Evaluator.

peastman:
The unified API looks good.  There are a lot of changes to type annotations.  I can't tell why you made them, or what they have to do with this PR.  Can you comment on them?

deepchem/metrics/metric.py
```diff
@@ -443,8 +443,8 @@ class Metric(object):
   """
 
   def __init__(self,
-               metric: Callable[..., float],
-               task_averager: Optional[Callable[..., Any]] = None,
```
nd-02110114, line 447: I added this type annotation, but  it was a mistake. So, I fixed.

deepchem/hyper/gaussian_process.py
nd-02110114: I just copied `GridHyperparamOpt` implementation

https://github.com/deepchem/deepchem/blob/02fce35729826b1ef12a1cfa6519b491510217be/deepchem/hyper/grid_search.py#L156-L157

deepchem/utils/evaluate.py
```diff
@@ -11,12 +11,10 @@
 
 logger = logging.getLogger(__name__)
 
-Score = Dict[str, float]
-Metric_Func = Callable[..., Any]
-Metrics = Union[Metric, Metric_Func, List[Metric], List[Metric_Func]]
```
nd-02110114, line 16: This annotation is difficult to read in documentation.

スクリーンショット 2020-10-21 16 11 54


So, I simplified the annotation and it is more readable annotation.

deepchem/feat/base_classes.py
```diff
@@ -233,7 +233,7 @@ class MolecularFeaturizer(Featurizer):
   The subclasses of this class require RDKit to be installed.
   """
 
-  def featurize(self, molecules, log_every_n=1000):
```
nd-02110114, line 236: The return type is missing, so I added

nd-02110114:
I reverted the type annotation modification which is not related to this PR. Sorry for your inconvenient.

deepchem/hyper/gaussian_process.py
peastman: What I mean is that all you need to do is add one argument to the existing call:

```python
multitask_scores = model.evaluate(valid_dataset, [metric], output_transformers)
```

nd-02110114: OK. I fixed.

nd-02110114:
The CI failed by gat tests, so I think we will merge in if @peastman approve

peastman:
Ok!

nd-02110114:
I merge in!

PR #2236, opened 2020-10-21
Fix graph conv model save/load

hsjang001205:
Most simply way to solve this issues.

#2237 should be discussed.


rbharath (approved):
@hsjang001205 Awesome debugging!! I was beating my head against this one for a few hours last night and couldn't come up with anything. This is a slick and elegant fix :)

PR #2240, opened 2020-10-22
Add more iterations for GAT test 

nd-02110114:
Recently, the gat test often fail. So, I fix the gat test by adding more iterations for traininga.


rbharath (approved):
Looks good! Feel free to merge once ready :)

nd-02110114:
Thank you for reviews! But, I think this modification is not needed because Mufei mentioned he will refactor the GAT model and I think his modification will resolve this slow convergence (see: https://github.com/deepchem/deepchem/pull/2249#issuecomment-717891737)

PR #2242, opened 2020-10-22
Fix bug for Mol2VecFingerprint

nd-02110114:
I found the bug about the usage of mol2vec function,  so I fixed.

### DIFF
--- a/deepchem/feat/molecule_featurizers/mol2vec_fingerprint.py
+++ b/deepchem/feat/molecule_featurizers/mol2vec_fingerprint.py
@@ -42,8 +42,7 @@ class Mol2VecFingerprint(MolecularFeaturizer):
   def __init__(self,
                pretrain_model_path: Optional[str] = None,
                radius: int = 1,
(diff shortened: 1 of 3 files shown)


peastman:
What bug is this fixing?

nd-02110114:
We should pass the List for the first argument of `sentences2vec` function, but I didn't pass. 
Therefore, the return value of `sentences2vec` was not a vector (N×M matrix) and I misunderstood that we should apply the some aggregation method like sum or mean to comvert the  N×M matrix to a M dim vector.

nd-02110114:

The CI failed by test_weave_fit_simple_infinity_distance, which is not related to this PR.
Recently, the CI is broken by test_weave_fit_simple_infinity_distance

rbharath (approved):
This looks good from my pass!

@nd-02110114 You know this featurizator much better than I do, so feel free to merge in once you feel this is ready :)

nd-02110114:
Thanks you for checking! I merge in!

PR #2243, opened 2020-10-22
Implement MACCS key and PubChem fingerprint featurizer

nd-02110114:
This PR adds two featurizers which create the fingerprint based on structural keys. 

The MACCS key, one of these, has been commonly used since before ECFP in the field of cheminformatics, so I think it is worth adding it to DeepChem. And then, I'm  joining the competition which predicts the smell from SMILES, and these fingerprints were more useful compared with ECFP or GNN variants. So, I submitted this PR.

MACCS key is a 166 bit structure key. 
PubChem Fingerprint is a 881 bit structure key and it is used by structure search in PubChem.


nd-02110114:

The CI failed  by `test_weave_fit_simple_infinity_distance`, which is not related to this PR.
Recently, the CI is broken by `test_weave_fit_simple_infinity_distance`

rbharath (approved):
Looks good! Feel free to merge once ready :)

nd-02110114:
Thank you for reviews! I merge in.

PR #2244, opened 2020-10-22

peastman:
Here is the next batch of loaders.  Some where straightforward to convert, but others involved complications.

The new API expects the list of tasks to be defined in the source code and passed to the `_MolnetLoader` object's constructor.   Clintox and Toxcast were instead written to load the tasks from the file, which meant you couldn't determine them until after you had downloaded and parsed the raw data.  Clintox was easy to convert since it only has two tasks, but Toxcast has a huge list of tasks.  Still, embedding that list in the source code seemed like the cleanest approach overall.

Zinc15 included a feature that isn't supported by the new API: it had a `splitter_kwargs` option that defined extra arguments to pass to `train_valid_test_split()` (*not* to the splitter's constructor, which the new API does support).  Unfortunately, it didn't work correctly.  The passed arguments didn't get recorded with cached datasets, so if you called it twice with two different sets of arguments, they would be ignored the second time and it would return datasets split with the original set of arguments.

The test case for the loader function relied on this feature to work.  It also didn't actually load the real dataset.  Instead it had a tiny data file and tried to trick the loader into thinking it had already downloaded the data and that was it.  Given these problems, I decided it was best to just comment out the test case.  We can test it properly as part of the larger project of figuring out how to test molnet loaders.

### DIFF
--- a/deepchem/molnet/load_function/zinc15_datasets.py
+++ b/deepchem/molnet/load_function/zinc15_datasets.py
@@ -2,59 +2,68 @@
 ZINC15 commercially-available compounds for virtual screening.
 """
 import os
-import logging
-import numpy as np
-import deepchem
-from deepchem.feat import Featurizer
-from deepchem.trans import Transformer
-from deepchem.splits.splitters import Splitter
-from deepchem.molnet.defaults import get_defaults
+import deepchem as dc
+from deepchem.molnet.load_function.molnet_loader import TransformerGenerator, _MolnetLoader
+from deepchem.data import Dataset
+from typing import List, Optional, Tuple, Union
 
-from typing import List, Tuple, Dict, Optional, Union
+ZINC15_URL = "https://deepchemdata.s3-us-west-1.amazonaws.com/datasets/"
+ZINC15_TASKS = ['mwt', 'logp', 'reactive']
 
-logger = logging.getLogger(__name__)
 
-DEFAULT_DIR = deepchem.utils.data_utils.get_data_dir()
+class _Zinc15Loader(_MolnetLoader):
 
-# dict of accepted featurizers for this dataset
-DEFAULT_FEATURIZERS = get_defaults("feat")
+  def __init__(self, *args, dataset_size: str, dataset_dimension: str,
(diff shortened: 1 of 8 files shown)


peastman:
Anyone have comments on these changes?

peastman:
Several tests in test_rdkit_descriptors.py are failing.  It looks like RDKitGridFeaturizer has started returning 208 features instead of 200?  We also have a failure in test_gat_classification that may just be an overly strict tolerance:

```
>     assert scores['mean-roc_auc_score'] >= 0.85
E     assert 0.8499053030303031 >= 0.85
```

And of course test_weave_fit_simple_infinity_distance is still failing.

rbharath (commented):
Did a brief pass over the changes and things look good overall. I was unfortunately not able to to do a thorough review (still can't really code because of my neck), but you should feel free to merge if you think the PR is in a good state!

I'm not sure what's up with the weave-infinity tests and other failures. It was passing until recently but suddenly stopped working last week. I'll try to take a look next week (if no one else takes a crack before then)

nd-02110114:
FYI: I fixed `test_rdkit_descriptors .py` in #2252 . And  the error of GAT test  is a stochastic error.

deepchem/molnet/load_function/zinc15_datasets.py
```diff
@@ -2,59 +2,68 @@
...
+        id_field="zinc_id",
+        featurizer=self.featurizer)
+    return loader.create_dataset(dataset_file, shard_size=8192)
 
 
 def load_zinc15(
-    featurizer=DEFAULT_FEATURIZERS['OneHotFeaturizer'],
-    transformers: List = [DEFAULT_TRANSFORMERS['NormalizationTransformer']],
-    splitter=DEFAULT_SPLITTERS['RandomSplitter'],
+    featurizer: Union[dc.feat.Featurizer, str] = dc.feat.OneHotFeaturizer(),
```
nd-02110114, line 57: This line also requires RDKit as a hard requirement and leads to the failure of importing deepchem.

PR #2252, opened 2020-10-27
Fix test which doesn't pass in the master branch and add some options for rdkit descriptor featurizer

nd-02110114:
The current matser build fails with RDKit descriptors and Weave tests. 
This PR fix the RDKit descriptor test and improve the RDKitFeaturizer.

I improve the following two things.

- Add use_fragment option.
  - RDKit descriptors contains numerical features and binary features, 
so sometimes we want to treat only numerical features for kernel base model or MLP.
- Add avg=True option for the IPC descriptor calculation
  - This option is needed to avoid IPC value is too large value
  - See this issue : https://github.com/rdkit/rdkit/issues/1527

### DIFF
--- a/deepchem/feat/molecule_featurizers/rdkit_descriptors.py
+++ b/deepchem/feat/molecule_featurizers/rdkit_descriptors.py
@@ -23,15 +23,29 @@ class RDKitDescriptors(MolecularFeaturizer):
   This class requires RDKit to be installed.
   """
 
-  def __init__(self):
+  def __init__(self, use_fragment=True, ipc_avg=True):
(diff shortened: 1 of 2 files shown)


nd-02110114:

The CI failed by test_weave_fit_simple_infinity_distance, which is not related to this PR.
Recently, the CI is broken by test_weave_fit_simple_infinity_distance

rbharath (approved):
Thanks for the fix here! I did a pass and I think this looks fine. Feel free to merge in once ready :)

nd-02110114:
Thanks! I merge in!

PR #2272, opened 2020-11-03
Fix datatype errors in model reload tests

ncfrey:
This should fix the errors popping up in #2269. I got rid of the `tf.keras.backend.set_floatx('float64')` statement in `test_normalizing_flow_model_reload`, which was causing errors in the other reload tests.


rbharath (approved):
Thanks for the fix!

PR #2275, opened 2020-11-03

peastman:
Here's the latest batch of loader functions.  All these ones were very straightforward, so I don't expect there to be many issues.  The one problem I ran into was switching the NCI dataset to use scaffold splitting.  That led to an exception:

> rdkit.Chem.rdchem.AtomValenceException: Explicit valence for atom # 0 Sn greater than permitted

For the moment I've left it using random splitting by default.  We should figure out what's causing the problem.


peastman:
The test failure is a doctest that fit a model on SAMPL, then expected the training set r2 to be > 0.9 and the validation set r2 to be > 0.7.  That only worked because it was using random splitting.  Now that it uses scaffold splitting, the training set score is still > 0.9, but the validation set score is only about 0.2.  I could just change the assertion, but that's pretty terrible model performance.  It might be better to change the example to a different dataset that does better.

rbharath (commented):
I did a quick pass and changes looks good!

For the SAMPL case, how about swapping it to one of the other regression datasets, perhaps Delaney, that's easier?

peastman:
I switched it to Delaney and tweaked the doctest a bit more.  It's now passing.

rbharath (approved):
Looks good! Feel free to merge in once ready :)

deepchem/molnet/load_function/clearance_datasets.py
```diff
@@ -2,115 +2,61 @@
...
+    featurizer: Union[dc.feat.Featurizer, str] = 'ECFP',
+    splitter: Union[dc.splits.Splitter, str, None] = 'scaffold',
+    transformers: List[Union[TransformerGenerator, str]] = ['normalization'],
+    reload: bool = True,
+    data_dir: Optional[str] = None,
+    save_dir: Optional[str] = None,
+    **kwargs
+) -> Tuple[List[str], Tuple[Dataset, ...], List[dc.trans.Transformer]]:
+  """
+  Load clearance datasets.
```
nd-02110114, line 36: Can you add some descriptions for this dataset?

deepchem/molnet/load_function/nci_datasets.py
```diff
@@ -4,119 +4,69 @@
...
+def load_nci(
+    featurizer: Union[dc.feat.Featurizer, str] = 'ECFP',
+    splitter: Union[dc.splits.Splitter, str, None] = 'random',
+    transformers: List[Union[TransformerGenerator, str]] = ['normalization'],
+    reload: bool = True,
+    data_dir: Optional[str] = None,
+    save_dir: Optional[str] = None,
+    **kwargs
+) -> Tuple[List[str], Tuple[Dataset, ...], List[dc.trans.Transformer]]:
+  """Load NCI dataset.
```
nd-02110114, line 46: Can you add some descriptions for this dataset?

deepchem/molnet/load_function/ppb_datasets.py
```diff
@@ -2,108 +2,59 @@
...
+def load_ppb(
+    featurizer: Union[dc.feat.Featurizer, str] = 'ECFP',
+    splitter: Union[dc.splits.Splitter, str, None] = 'scaffold',
+    transformers: List[Union[TransformerGenerator, str]] = ['normalization'],
+    reload: bool = True,
+    data_dir: Optional[str] = None,
+    save_dir: Optional[str] = None,
+    **kwargs
+) -> Tuple[List[str], Tuple[Dataset, ...], List[dc.trans.Transformer]]:
+  """Load PPB datasets.
```
nd-02110114, line 34: The same here

nd-02110114 (commented):
I commented about only documents. Thanks you for updating API!

peastman:
@nd-02110114 Can you open a separate issue for the documentation problems?  This PR is just code changes.  I don't know what the origins of those datasets are, but we should track them down and document them!

PR #2300, opened 2020-11-16
Flake8 fix in vina_utils.py

ncfrey:
There was a trailing space in `vina_utils.py` introduced in #2298 that was causing travis failures, as mentioned in #2297. Should be fixed now.


ncfrey:
It looks like all checks have passed so I'll go ahead and merge this in to fix the failing tests.

PR #2313, opened 2020-12-04
Fixed errors installing PyTorch Geometric

peastman:
I think this was the cause of one of the test failures that just started appearing on travis.  I changed the install scripts to match the instructions at https://pytorch-geometric.readthedocs.io/en/latest/notes/installation.html.  Perhaps they just changed how they label the packages?


peastman:
This fixed the problem.  We still have the failure in `test_tokenize()`, but that's something different.

rbharath (approved):
Thanks for the fix! Feel free to merge once ready :)

PR #2319, opened 2020-12-12
changing the expected output of get_task_names for DiskDataset to List[str]

HFooladi:
It is a very very minor correction to the `DiskDataset` object. `DiskDataset` object has `get_task_names()` method and I believe the expected output for this method is `List[str]` instead of `np.ndarray`. So, I changed this one line.


rbharath (approved):
Looks good! Thanks for the fix here :). The test failures are unrelated (and due to some other issues with our tokenizers so I'm going to merge in. Congrats on your first accepted DeepChem PR!

Issue #2323, opened 2020-12-13
RuntimeError: 1D target tensor expected, multi-target not supported


mufeili:
The error is located at [L203, SparseSoftmaxCrossEntropy](https://github.com/deepchem/deepchem/blob/master/deepchem/models/losses.py#L203), where `output` is of shape `(N, C)` and `labels` is of shape `(N, 1)`. In this case, `CrossEntropyLoss` expects `labels` to have shape `(N)` and we may fix that with `labels.squeeze(-1)`. Meanwhile, this loss function is supposed to support multitask as well and I'm not sure what will be the shape of `labels` in that case. Any suggestions @nd-02110114 ?

nd-02110114:
In my understanding,
in the case of single task, SparseSoftmaxCrossEntropy expects 
labels have shape `(batch_size)` and outputs have shape `(batch_size, num_classes)`.
in the case of multi task, SparseSoftmaxCrossEntropy expects 
labels have shape `(batch_size, num_tasks)` and outputs have shape `(batch_size, num_tasks, num_classes)`.

mufeili:
When I tried fitting a GCNModel on BACE, I encountered this error. It seems that `labels[0]` [here](https://github.com/deepchem/deepchem/blob/master/deepchem/models/torch_models/gcn.py#L352) is of shape `(batch_size, 1)`. Is this expected? If not, should I modify the code for `GCNModel` or `TorchModel._prepare_batch`?

peastman:
Can you post an example that produces an error?

mufeili:
> Can you post an example that produces an error?

See below. You may need to use the latest code in the master branch.

```python
import deepchem as dc
import torch.nn.functional as F
from deepchem.feat import MolGraphConvFeaturizer
from deepchem.molnet import load_bace_classification

tasks, all_dataset, transformers = load_bace_classification(featurizer=MolGraphConvFeaturizer(), splitter='scaffold')
train_set, val_set, test_set = all_dataset
model = dc.models.GCNModel(
            n_tasks=len(tasks),
            graph_conv_layers=[32],
            activation=F.relu,
            residual=False,
            batchnorm=False,
            dropout=0.,
            predictor_hidden_feats=32,
            predictor_dropout=0.,
            mode='classification',
            n_classes=2,
            learning_rate=3e-4)
model.fit(train_set, nb_epoch=1)
```

nd-02110114:
I understood the situation. I think there are two options.

[...]

PR #2328, opened 2020-12-16
Reshape Labels for Single-Task Datasets for torch.nn.CrossEntropyLoss

mufeili:
# Pull Request Template

## Description

As described in #2323 , `CrossEntropyLoss` expects the labels to have shape `(N)` for single-task cases, where `N` is the number of datapoints. We decide to reshape the labels for ~~the model wrappers.~~ the loss wrappers.

Can you review it? @nd-02110114 

`mypy` seems to yield errors irrelevant to this PR.

```bash
deepchem/utils/fragment_utils.py:194: error: Need type annotation for 'all_atoms' (hint: "all_atoms: List[<type>] = ...")
deepchem/data/datasets.py:2917: error: Need type annotation for 'datasets' (hint: "datasets: Dict[<type>, <type>] = ...")
deepchem/feat/molecule_featurizers/mol_graph_conv_featurizer.py:203: error: Need type annotation for 'src' (hint: "src: List[<type>] = ...")
deepchem/feat/molecule_featurizers/mol_graph_conv_featurizer.py:203: error: Need type annotation for 'dest' (hint: "dest: List[<type>] = ...")
deepchem/feat/molecule_featurizers/mol_graph_conv_featurizer.py:213: error: Need type annotation for 'bond_features' (hint: "bond_features: List[<type>] = ...")
deepchem/feat/graph_features.py:619: error: Need type annotation for 'next_adj' (hint: "next_adj: Set[<type>] = ...")
deepchem/data/data_loader.py:912: error: Need type annotation for 'remainder' (hint: "remainder: List[<type>] = ...")
deepchem/splits/splitters.py:563: error: Need type annotation for 'assigned' (hint: "assigned: Set[<type>] = ...")
deepchem/splits/splitters.py:1262: error: Need type annotation for 'scaffolds' (hint: "scaffolds: Dict[<type>, <type>] = ...")
deepchem/splits/splitters.py:1461: error: Need type annotation for 'years' (hint: "years: Dict[<type>, <type>] = ...")
deepchem/models/torch_models/torch_model.py:200: error: Need type annotation for '_prediction_outputs' (hint: "_prediction_outputs: List[<type>] = ...")
deepchem/models/torch_models/torch_model.py:201: error: Need type annotation for '_loss_outputs' (hint: "_loss_outputs: List[<type>] = ...")
deepchem/models/torch_models/torch_model.py:202: error: Need type annotation for '_variance_outputs' (hint: "_variance_outputs: List[<type>] = ...")
deepchem/models/torch_models/torch_model.py:203: error: Need type annotation for '_other_outputs' (hint: "_other_outputs: List[<type>] = ...")
deepchem/models/torch_models/torch_model.py:542: error: Need type annotation for 'access_values' (hint: "access_values: List[<type>] = ...")
deepchem/models/keras_model.py:208: error: Need type annotation for '_prediction_outputs' (hint: "_prediction_outputs: List[<type>] = ...")
deepchem/models/keras_model.py:209: error: Need type annotation for '_loss_outputs' (hint: "_loss_outputs: List[<type>] = ...")
deepchem/models/keras_model.py:210: error: Need type annotation for '_variance_outputs' (hint: "_variance_outputs: List[<type>] = ...")
deepchem/models/keras_model.py:211: error: Need type annotation for '_other_outputs' (hint: "_other_outputs: List[<type>] = ...")
deepchem/models/keras_model.py:627: error: Need type annotation for 'access_values' (hint: "access_values: List[<type>] = ...")
deepchem/dock/pose_generation.py:274: error: Need type annotation for 'docked_complexes' (hint: "docked_complexes: List[<type>] = ...")
deepchem/dock/pose_generation.py:275: error: Need type annotation for 'all_scores' (hint: "all_scores: List[<type>] = ...")
Found 22 errors in 9 files (checked 299 source files)
```

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/losses.py
+++ b/deepchem/models/losses.py
@@ -200,6 +200,8 @@ class SparseSoftmaxCrossEntropy(Loss):
       # This is for API consistency
       if len(output.shape) == 3:
         output = output.permute(0, 2, 1)
+      if labels.shape[-1] == 1:
+        labels = labels.squeeze(-1)
       return ce_loss(output, labels.long())
 
     return loss
--- a/deepchem/models/losses.py
+++ b/deepchem/models/losses.py
@@ -188,7 +188,7 @@ class SparseSoftmaxCrossEntropy(Loss):
   def _compute_tf_loss(self, output, labels):
     import tensorflow as tf
 
-    if labels.shape[-1] == 1:
+    if len(labels.shape) == len(output.shape):
       labels = tf.squeeze(labels, axis=-1)
 
     labels = tf.cast(labels, tf.int32)
@@ -205,7 +205,8 @@ class SparseSoftmaxCrossEntropy(Loss):
       # This is for API consistency
       if len(output.shape) == 3:
         output = output.permute(0, 2, 1)
-      if labels.shape[-1] == 1:
+
+      if len(labels.shape) == len(output.shape):
         labels = labels.squeeze(-1)
       return ce_loss(output, labels.long())
 
--- a/deepchem/models/losses.py
+++ b/deepchem/models/losses.py
@@ -187,7 +187,12 @@ class SparseSoftmaxCrossEntropy(Loss):
 
   def _compute_tf_loss(self, output, labels):
     import tensorflow as tf
+
+    if labels.shape[-1] == 1:
+      labels = tf.squeeze(labels, axis=-1)
+
     labels = tf.cast(labels, tf.int32)
+
     return tf.nn.sparse_softmax_cross_entropy_with_logits(labels, output)
 
   def _create_pytorch_loss(self):
--- a/deepchem/models/tests/test_losses.py
+++ b/deepchem/models/tests/test_losses.py
@@ -186,6 +186,12 @@ class TestLosses(unittest.TestCase):
     expected = [-np.log(softmax[0, 1]), -np.log(softmax[1, 0])]
     assert np.allclose(expected, result)
 
+    labels = tf.constant([[1], [0]])
+    result = loss._compute_tf_loss(outputs, labels).numpy()
+    softmax = np.exp(y) / np.expand_dims(np.sum(np.exp(y), axis=1), 1)
+    expected = [-np.log(softmax[0, 1]), -np.log(softmax[1, 0])]
+    assert np.allclose(expected, result)
+
   @unittest.skipIf(not has_pytorch, 'PyTorch is not installed')
   def test_sparse_softmax_cross_entropy_pytorch(self):
     """Test SparseSoftmaxCrossEntropy."""
--- a/deepchem/models/tests/test_losses.py
+++ b/deepchem/models/tests/test_losses.py
@@ -198,6 +198,12 @@ class TestLosses(unittest.TestCase):
     expected = [-np.log(softmax[0, 1]), -np.log(softmax[1, 0])]
     assert np.allclose(expected, result)
 
+    labels = torch.tensor([[1, 0]])
+    result = loss._create_pytorch_loss()(outputs, labels).numpy()
+    softmax = np.exp(y) / np.expand_dims(np.sum(np.exp(y), axis=1), 1)
+    expected = [-np.log(softmax[0, 1]), -np.log(softmax[1, 0])]
+    assert np.allclose(expected, result)
+
   @unittest.skipIf(not has_tensorflow, 'TensorFlow is not installed')
   def test_VAE_ELBO_tf(self):
     """."""
--- a/deepchem/models/tests/test_losses.py
+++ b/deepchem/models/tests/test_losses.py
@@ -198,7 +198,7 @@ class TestLosses(unittest.TestCase):
     expected = [-np.log(softmax[0, 1]), -np.log(softmax[1, 0])]
     assert np.allclose(expected, result)
 
-    labels = torch.tensor([[1, 0]])
+    labels = torch.tensor([[1], [0]])
     result = loss._create_pytorch_loss()(outputs, labels).numpy()
     softmax = np.exp(y) / np.expand_dims(np.sum(np.exp(y), axis=1), 1)
     expected = [-np.log(softmax[0, 1]), -np.log(softmax[1, 0])]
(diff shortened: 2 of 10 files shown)


peastman:
I think it would be better to make the change in `SparseSoftmaxCrossEntropy`.  Here is the documented behavior of that class:

>   The labels should have shape (batch_size) or (batch_size, tasks), and be
  integer class labels.  The outputs have shape (batch_size, classes) or
  (batch_size, tasks, classes) and be logits that are converted to probabilities
  using a softmax function.

If it isn't behaving that way, it's a bug.  Since it's implemented using `torch.nn.CrossEntropyLoss`, it needs to reshape its inputs as necessary to satisfy the needs of that class.  Something like this:

```python
if len(outputs.shape) == len(labels.shape) and labels.shape[1] == 1:
  labels = labels.squeeze(1)
```

nd-02110114:
The mypy errors you shows doesn't happen in the CI and I also can't reproduce. 

> If it isn't behaving that way, it's a bug.

I  think the loss function behaves correctly.
The problems is that user often use the labels have shape (batch_size, 1) although the labels must have shape (batch_size,) for a single task classification. 

Actually, all molnet functions return the labels have shape (N, 1) for a single task classification. So, whenever some users use Models which have `SparseSoftmaxCrossEntropy ` with molnet functions, they need to reshape the labels. Mufei point out it is troblesome.

peastman:
>  The problem is that user often use the labels have shape (batch_size, 1) although the labels must have shape (batch_size,) for a single task classification.

According to the docstring, (batch_size, 1) is also a legal shape for the input.  Also, the TensorFlow implementation works with inputs of that size.  It's just the PyTorch version that doesn't, and that's a bug.

nd-02110114:
From the docstring, I assumed the loss function supports one to one relationship like the following.

labels `(batch_size)` <-> outputs `(batch_size, classes)`
labels `(batch_size, tasks)` <-> outputs `(batch_size, tasks, classes)`

But, there are no specific description about one to one relationship above. So, I could understand why you think it is bug. I'm not native in English, so I'm so sorry for your troublesome.

> Also, the TensorFlow implementation works with inputs of that size. It's just the PyTorch version that doesn't, and that's a bug.

OK. So, we should modify the loss functions directly and add more test cases.

peastman:
> From the docstring, I assumed the loss function supports one to one relationship like the following.

Well, that wasn't what I intended when I wrote it. 😀  And that would certainly be inconvenient since, as you pointed out, the molnet loaders produce labels of shape (samples, 1).

mufeili:
@peastman @nd-02110114 I've made the update for the loss wrapper.

rbharath (commented):
Just had a chance to look at this PR. Changes look reasonable! Will defer to @peastman @nd-02110114 on when this is ready to merge in :)

deepchem/models/tests/test_losses.py
```diff
@@ -198,6 +198,12 @@ def test_sparse_softmax_cross_entropy_pytorch(self):
     expected = [-np.log(softmax[0, 1]), -np.log(softmax[1, 0])]
     assert np.allclose(expected, result)
 
+    labels = torch.tensor([[1], [0]])
```
nd-02110114, line 201: Can you add the same test for TensorFlow?

nd-02110114 (commented):
I reviewed! Basically, it looks good to me.

deepchem/models/tests/test_losses.py
mufeili: Done.

peastman:
The tests don't have any assertions.  Are they just making sure it doesn't crash?

mufeili:
> The tests don't have any assertions. Are they just making sure it doesn't crash?

Yes.

peastman:
Ok.  Looks like you have a test failure in `test_sparse_softmax_cross_entropy_tf()`.

mufeili:
> Ok. Looks like you have a test failure in `test_sparse_softmax_cross_entropy_tf()`.

This should be fixed now.

deepchem/models/losses.py
```diff
@@ -187,7 +187,12 @@ class SparseSoftmaxCrossEntropy(Loss):
 
   def _compute_tf_loss(self, output, labels):
     import tensorflow as tf
+
+    if labels.shape[-1] == 1:
+      labels = tf.squeeze(labels, axis=-1)
```
peastman, line 192: I don't think that's the right check for TensorFlow.  `sparse_softmax_cross_entropy_with_logits()` requires `labels` to have one fewer dimension than `output`.  We should check whether `len(labels.shape) == len(output.shape)+1`.

mufeili: How about adding an additional assertion statement after L192 for the check you proposed? There can be multiple reasons for this. L191/192 are mainly handling the issue in molnet data loader.

peastman: It's not an assertion.  The issue is that this function is more permissive of its inputs than the TensorFlow and PyTorch functions we use to implement it.  Given any inputs that are legal for this function, we need to reshape them as required for the TensorFlow and PyTorch functions.  Conversely, we should only reshape them if the inputs are *not* the correct shape for the implementing function.  If the inputs already have the shape expected by `sparse_softmax_cross_entropy_with_logits()`, we shouldn't call `squeeze()`.

mufeili: What check/reshape will you suggest? It's not guaranteed that one can obtain `len(labels.shape) == len(output.shape)+1` by reshaping as the inputs may not be valid themselves.

peastman: If the inputs aren't valid, that's the user's problem, not ours. 😀  How about

```python
if len(labels.shape) == len(outputs.shape):
  labels = tf.squeeze(labels, axis=-1) 
```

mufeili: Sounds good. I've made the update.

mufeili: With "thumbs up", I assume it looks good to you 😄 . Any further issues? @peastman

peastman: I think it's good to merge.

mufeili:
@rbharath You may go ahead and merge it. The failed tests are related to DTNN,

Issue #2248, opened 2020-10-26
Do we need validation dataset in random forest?


hjzhang1018:
Hi, I'm trying to understand and apply hyperparameter tuning in Random Forest model. However I saw a lot o f tutorials did not use validation dataset in these parameter tuning steps. Like this snippet shown below,
```
param_grid = {"n_estimators": range(50, 200, 50),
               "max_features": ["auto", "sqrt", "log2", None]}
rf_grid = GridSearchCV(estimator=RandomForestClassifier(random_state=10),
                       param_grid=param_grid,
                       scoring="roc_auc",
                       cv=5)
rf_grid.fit(xtrain, ytrain)
print("rf_grid best parameters: %s" % str(rf_grid.best_params_))
print("rf_grid best score: %s" % str(rf_grid.best_score_))
```
Looks like we're still using training dataset for a better score. Correct me if I'm wrong, I wonder in other models, like GCNN, do we also not use the validation dataset for parameter tuning ? Thanks a lot for any advice!

peastman:
Hyperparameter tuning should always be done with the validation set.  Which tutorials aren't doing that?

hjzhang1018:
Yes I  agree @peastman . I understand that validation data are used for this. In deepchem `hyperparam_search`, I know we passed both `train dataset` and `valid dataset` to this searching steps, but in this snippet I didn’t see where we passed the validation dataset to the grid searching steps. So I'm a bit confused.

peastman:
Where did it come from?

hjzhang1018:
It comes from a website "https://www.kaggle.com/funxexcel/p2-sklearn-randon-forest-parameter-tuning". 
Could you please tell me how to pass the valid dataset to the tuning steps in this case? Thank you very much! @peastman

peastman:
That website isn't anything related to DeepChem.  For a description of how to do hyperparameter optimization in DeepChem, see this tutorial: https://github.com/deepchem/deepchem/blob/master/examples/tutorials/09_Advanced_Model_Training.ipynb

hjzhang1018:
Thank you @peastman! I learned this tutorial and this is the snippet I used for random forest hyperparameter tuning. However it shows an error message and I didn’t figure a way to solve it. Could you help me with this bug? Thank you very much.
```
def rf_model_builder(model_params, model_dir):
  sklearn_model = RandomForestClassifier(**model_params)
  n_estimators = model_params["n_estimators"]
[...]

PR #2349, opened 2021-01-11
Fixed incorrect sample code

peastman:
See https://github.com/deepchem/deepchem/issues/2248#issuecomment-757077199.


peastman:
The test failures are unrelated to this.

PR #2406, opened 2021-02-22
[WIP] AtomicConvFeaturizer test fixes

ncfrey:
## Description

Partial fix for #2373

This PR fixes the errors in `test_atomic_conv_feat.py`. The errors were caused by the number of atoms in the protein and complex changing during featurization because of the `_strip_hydrogens` step. The coordinates list and Z-matrix features are padded to match the number of atoms specified in the input, while the neighbor list is not, which was causing the errors. 

I've uncommented the failing tests, fixed them, and added a few more explicit tests to make sure that the featurizer is generating non-zero outputs.

## Type of change


rbharath (commented):
This looks great so far! Good catch on the number of atoms changing being the root of the issue with the failure

ncfrey:
Getting the same `torch` install errors here that have popped up in recent PRs. Looks like it's related to the new release of [pytorch spline conv](https://github.com/rusty1s/pytorch_spline_conv/releases/tag/1.2.1)

ncfrey:
`models.atomic_conv.py` will show up as a changed file, but that is only reformatting that was required because the file was altered in previous commits. I will address fixes for the actual model and model tests in a future PR.

ncfrey:
Flaky tests are failing but `AtomicConvFeaturizer` tests are now passing!
CC @rbharath

rbharath (approved):
Changes to `AtomicConvFeaturizer` look good! I think this PR should be good to merge ince once you're ready

ncfrey:
Going to go ahead and merge this in and move on to fixing the slow tests for `AtomicConvModel`!

PR #2408, opened 2021-02-23
Fixing flaky test

sleepy-owl:
Hi,

The test `test_uncertainty` sometimes fails non-deterministically. The assertion `assert (np.mean(std) < 1.0)` fails sometimes when np.mean(std) exceeds 1.0.

To find a solution, I collected several samples of `np.mean(std)` from several test executions and computed the tail distribution. I computed the extreme percentiles to check how low can the values be:

```
0.9 :: 0.63
0.99 :: 0.95
0.9999 :: 1.58 ~ 1.6
```
It seems setting this to 1.6 might be good enough to fix this test.  I think setting the bound using the statistical evaluation might be a good way to ensure the test is not flaky.

Do you guys think this makes sense? Please let me know if this looks good or if you have any other suggestions. Also, here I assume there are no bugs in the code under test.

Thanks!

### DIFF
--- a/deepchem/models/tests/test_kerasmodel.py
+++ b/deepchem/models/tests/test_kerasmodel.py
@@ -227,7 +227,7 @@ def test_uncertainty():
   model.fit(dataset, nb_epoch=2500)
   pred, std = model.predict_uncertainty(dataset)
   assert np.mean(np.abs(y - pred)) < 1.0
-  assert noise < np.mean(std) < 1.0
+  assert noise < np.mean(std) < 1.6
 
 
 def test_saliency_mapping():


peastman:
It would be good to understand this behavior a little better.  We can always make a test pass by increasing the tolerances far enough.  The question is, what goes wrong that causes it to be that far off?  We create the training data like this:

https://github.com/deepchem/deepchem/blob/458b2911b7d0dcd495e53878e8329c9d36634152/deepchem/models/tests/test_kerasmodel.py#L177-L179

So the true standard deviation of the training data should be about 0.1.  Epistemic uncertainty increases it, so we require the estimated uncertainties to be >0.1, but it shouldn't be that much higher.  When it comes out >1, what's going on that causes that?  Is it because the training set randomly happened to have unusually high variation?  If so, maybe we should base the test on `np.std(y)`.  Or has the model just not been trained very well, and we should increase the number of epochs?

PR #2416, opened 2021-02-26
More concise code

aced125:
## Description

Why isn't the xyz extraction implemented like this?

I checked the two methods on Atorvastatin; both give the same result.


rbharath (commented):
I think we could indeed use this instead! Could you remove the commented lines and run yapf on the code? The docs at https://deepchem.readthedocs.io/en/latest/development_guide/coding.html should hopefully be a good guide on what's needed

PR #2426, opened 2021-03-06
Basic MolGAN model

MiloszGrabski:
# BasicMolGAN model

## Description

Final stage of implementation of basic version of [MolGAN](https://arxiv.org/abs/1805.11973) originally created by Nicola De Cao and Thomas Kipf.
This is a generative graph convolution model based on WGAN architecture that enables generation of small molecules.
Smiles are converted into model input by MolGANFeaturizer class (tested on MoleculeNet QM9 dataset).
Currently, training is a bit unstable and sometimes requires a few separate training attempts to generate a valid model.
This is something I hope to fix in the future release.
Jupyter notebook tutorial will be provided in a separate PR.

## Type of change

## Checklist


MiloszGrabski:
Issue not connected with my code.
===End Flaky Test Report===
=========================== short test summary info ===========================
FAILED deepchem/splits/tests/test_scaffold_splitter.py::TestScaffoldSplitter::test_scaffolds
= 1 failed, 688 passed, 6 skipped, 35 deselected, 695 warnings in 2406.46s (0:40:06) =

deepchem/models/molgan.py
```diff
@@ -0,0 +1,258 @@
...
+
+class BasicMolGANModel(WGAN):
+  """
+  Model for de-novo generation of small molecules based on work of Nicola De Cao et al. [1]_.
+  Utilizes WGAN infrastructure; uses adjacency matrix and node features as inputs.
+  Inputs need to be one-hot representation.
+
+  Examples
+  --------
+  import deepchem as dc
```
rbharath, line 19: Could you format this as follows:

```
Examples
--------
>>> import deepchem as dc
>>> from deepchem.imports import BasicMolGANModel as MolGAN
...
```

That will let the docs render the example code :)

deepchem/models/molgan.py
```diff
@@ -0,0 +1,258 @@
...
+  def create_generator(self) -> keras.Model:
+    """
+    Create generator model.
+    Take noise data as an input and processes it through number of
+    dense and dropout layers. Then data is converted into two forms
+    one used for training and other for generation of compounds.
+
+    Returns
+    -------
+    keras.Model
```
rbharath, line 123: I'm not sure if this will render correctly in the docs. Could you try building the docs locally and seeing if this looks good?

rbharath (commented):
Did a first round of review. Looks good overall! I've made a few comments about docs mainly below.

@peastman Could you do a review as well? This model would be the first molecular GAN in the library and we might want to make sure the API is one we like

deepchem/models/molgan.py
MiloszGrabski: There is problem with this as doctest fails due to not expecting values when model is being trained.
I have tried to solve it but no matter what I tried it did not work as training is variable.

deepchem/models/molgan.py
MiloszGrabski: Do you mean running doctest?

deepchem/models/molgan.py
peastman: If you want it to be formatted as code but not run as a doctest, you can do this:

```python
>>>
>> import deepchem as dc
>> from deepchem.imports import BasicMolGANModel as MolGAN
...
```

deepchem/models/tests/test_molgan_model.py
```diff
@@ -0,0 +1,70 @@
...
+
+from deepchem.models import BasicMolGANModel as MolGAN
+
+
+class test_molgan_model(unittest.TestCase):
+  """
+  Unit testing for MolGAN basic layers
+  """
+
+  def __init__(self, *args, **kwargs):
```
peastman, line 11: You should use a `setup()` method rather than a constructor for setting these.  That way it will be executed before each test, and you don't have to worry about information leaking between tests.

peastman:
Is it possible to add a test for correct operation?  The current tests just check that certain parameters were recorded correctly.  They don't tell you whether the model works.  For example, can you train a small model, have it predict some molecules, and verify that they're valid?

deepchem/models/tests/test_molgan_model.py
MiloszGrabski: could you show an example, so I understand what you mean?

MiloszGrabski:
Problem is that it not always trains model that can generate valid molecules.
Not sure how to accommodate that fact.

peastman:
Here is an example of a GAN test case:

https://github.com/deepchem/deepchem/blob/7ed0125d9ad26cc4cbf31fbe8f67c195448c71db/deepchem/models/tests/test_gan.py#L133-L148

It trains a GAN based on a very simple data distribution, and checks whether the generated data is reasonably close.  Can you do something similar for this class?  Train it on a very simple dataset, and see whether it has learned the basic properties of the training data?

MiloszGrabski:
I am not sure as it generates molecules and not always learns to do so.
But I have not tried on a very small amount of compounds.
I will check if this might work.

deepchem/models/tests/test_molgan_model.py
MiloszGrabski: @peastman I think I now understand what you mean, but to be sure, do you meant [this](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp)?

peastman: Right.  Here's an example from an existing test:

https://github.com/deepchem/deepchem/blob/c68f7bab307599f159a8896b63a7e7e39e46725c/deepchem/models/tests/test_pretrained_keras.py#L38-L46

deepchem/models/molgan.py
MiloszGrabski: Hi I tried running sphinx, but struggling to get it working.
Should I just remove Returns part?

MiloszGrabski:
I have run pytest on test_molgan_model and it worked fine. Not sure what is the issue here.

MiloszGrabski:
ok, so there seems to be corner case where there is a  conflict between yapf and flake8. 
Any idea about workaround?

peastman:
> ok, so there seems to be corner case where there is a conflict between yapf and flake8.

How do you mean?  It's just complaining about the indentation of one line.

On the other hand, there seems to be an actual test failure on Windows (but not Linux):

```
FAILED deepchem/models/tests/test_molgan_model.py::test_molgan_model::test_training
= 1 failed, 690 passed, 6 skipped, 35 deselected, 696 warnings in 2863.04s (0:47:43) =
Exception ignored in: <function Model.__del__ at 0x000001BFDC1B34C8>
  File "D:\a\deepchem\deepchem\deepchem\models\models.py", line 61, in __del__
    shutil.rmtree(self.model_dir)
  File "c:\miniconda3\envs\deepchem\lib\shutil.py", line 516, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "c:\miniconda3\envs\deepchem\lib\shutil.py", line 404, in _rmtree_unsafe
    onerror(os.rmdir, path, sys.exc_info())
  File "c:\miniconda3\envs\deepchem\lib\shutil.py", line 402, in _rmtree_unsafe
    os.rmdir(path)
OSError: [WinError 145] The directory is not empty: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpkauecabw'
```

peastman:
The new test case looks good.  Just out of curiosity, could you post a few examples of molecules it generates?  I'm just wondering what sorts of molecules it produces after training on that small number of examples.  Has it memorized members of the training set?  Or is it just outputting strings of 'C's?

MiloszGrabski:
Actually, I was surprised to see that it did not recreated single molecule from original set.
Some of them were odd looking some of them were fine.
There was single case so far that was just C's.
In terms of indentation, this is precisely the case when flask8 wants different indentation than yafp (searched in google and I am not the only one experiencing this issue).
So when I fixed it in flake8, yapf tripped over it. I will try to rewrite code somewhat. Maybe it is worth considering switching test to do flake8 and yapf separately. I do not have much experience with automatic code testing, so not sure if this is a good idea.

MiloszGrabski:
Not sure what is the issue here, I run pytest on my machine and it passed:
pytest -v -m "not slow" D:\deepchem\deepchem\models\tests\test_molgan_model.py
image

MiloszGrabski:
# Example model outputs (sample of 1000 generated each time):
The model was trained in the loop of 10, with each loop creating new model.
The loop breaks when valid molecules are generated, so first successful model is retained.
Luckily enough, for these attempts it generated models at first attempt.
As you can see, it has quite low validity/uniqueness (based on smiles) score.
The authors solved it by implementing a few things which I would like to add in the future release.
One of them, which seems the most straightforward, is guiding via model checking after each training session.
If model performs worse go back to previous checkpoint otherwise use current.
The infrastructure still needs a lot of work to be useful, but the base seems solid enough.

What I also noticed that with more compounds low number of iterations is favourable (8-10), whereas for small number (like in the test set) you need much larger number (1000-5000). 
I suppose it makes sense as with large sets you get much more exposure per training loop.
When I plotted losses, generally around 1000 was enough for them to stabilize.
But loss itself cannot be used as measure of model performance, because even with small losses the model 
can generate a lot of invalid molecules i.e. I have observed models with worse losses that provided better metrics.

### Model 1, 804 valid, 6 unique, first training attempt successful:
image
### Model 2, 60 valid, 10 unique, first training attempt successful:
image
### Model 3, 13 valid, 7  unique, first training attempt successful
image
### Model 4,  17 valid, 3 unique, first training attempt successful
image

peastman:
Those actually aren't too bad.  Each model produces a decent amount of diversity.  It does seem to have a tendency to produce the same molecules over and over, but that's not surprising after overfitting to a tiny dataset.

>  Luckily enough, for these attempts it generated models at first attempt.

If it usually produces valid molecules on the first attempt, maybe we should reduce the number of iterations?  Letting it fail nine times before succeeding could cause the test not to notice if a future change makes the code stop working as well.

rbharath:
These don't look bad at all! I'd be curious about how these compare to molecules generated by the normalizing flows. CC @ncfrey who might have some insight there

ncfrey:
Looks really cool @MiloszGrabski! The validity and uniqueness scores are pretty low, but that might not matter too much. About how long does it take to generate 1000 samples?

MiloszGrabski:
@ncfrey It is almost instantaneous, I would say 1 sec maybe. Never paid it much attention as it was so fast.
That is why in the future version I want to implement guiding system in form of selection which model progress to the next epoch.
This will require sample generation after each epoch/applying metric, which would be prohibitive if not fast enough.
@peastman I think though of what you said and will modify code slightly. Let me know what you think of the change.

MiloszGrabski:
Do not know what is going on here:
Works fine on my machine, any ideas?
image
Unittest also worked just fine.
From error it looks like input mismatch, but no idea why it is happening.
image
I would welcome any suggestion why it might be.

MiloszGrabski:
I think I know why the issue albeit do not understand why it works on my machine but does not on test.
My generator output is = [a,b,c,d]
My discriminator input is = [a,b]
Normally additional inputs are just ignored, but in the test case it seems to behave differently. Any clue?

rbharath:
One thought off hand is perhaps it's an issue with different randomness? Perhaps setting the random seed to the same value could help here.

MiloszGrabski:
The current errors seems to be connected with gan itself:
Windows test, line 847
@peastman  any clue?
       
 # Train the discriminator.
    
        inputs = [self.get_noise_batch(self.batch_size)]
        for input in self.data_input_layers:
>         inputs.append(feed_dict[input.ref()])
E         KeyError: <Reference wrapping <KerasTensor: shape=(None, 9, 9) dtype=float32 (created by layer 'input_4')>>

deepchem\models\gan.py:348: KeyError
============================== warnings summary ===============================
c:\miniconda3\envs\deepchem\lib\site-packages\tensorflow\python\autograph\impl\api.py:22
  c:\miniconda3\envs\deepchem\lib\site-packages\tensorflow\python\autograph\impl\api.py:22: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

MiloszGrabski:
I was thinking about creating simple layer that will decide which type of output to provide.
But I am struggling with finding a good method to pass signal into layers. 
Something similar to training, any ideas?
I am trying to add additional input, but so far without much luck.

peastman:
I don't think that design is going to work very well.  As the documentation for `create_generator()` says,

https://github.com/deepchem/deepchem/blob/ab5fa059174b24a9fe98f1c3b779d89b7a1eee67/deepchem/models/gan.py#L242-L244

If you violate that requirement, you're going to create all sorts of problems for yourself.

Can't you do everything you need just by overriding `predict_gan_generator()`?  Given the softmax outputs, it looks like it ought to be easy to calculate the argmax ones, and the only time you need them is when you're predicting molecules.

MiloszGrabski:
@peastman I was looking for a way but do not see how.
The problem is that real data goes through one-hot encoding, fake data  logits goes through softmax for discriminator or gumbel then argmax for sample generation.
I cannot simply return logits as then discriminator would have to handle conversion and differentiation between real/fake data.
The only option would to reverse softmax in order to get logits in predict_gan_generator(). Do you know any built-in method to do so?

MiloszGrabski:
@rbharath @peastman 
It seems the issues are not connected with PL
FAILED deepchem/feat/tests/test_mol2vec_fingerprint.py::TestMol2VecFingerprint::test_mol2vec_fingerprint
FAILED deepchem/models/tests/test_mpnn.py::test_mpnn_classification - assert ...

rbharath (approved):
My apologies for the slow review turnaround! I think this is good to merge on my end

Once @peastman is done with his review, we can go ahead and merge this in

rbharath:
Wanted to check in on this PR. I'll plan to go ahead and merge in tomorrow in case there's no further comments

MiloszGrabski:
Once you merge, I will do a few small tweaks and create tutorial in jupyter lab

rbharath:
@MiloszGrabski congrats on the merge! Major new feature added to DeepChem :). I'll be sure to give this a shout-out in the DeepChem 2.6 release notes (tentative for Mid May)

PARODBE:
@MiloszGrabski thank you for this library, could you create a notebook for all the options: molgan.py and test_molgan

Thank you!

Issue #2438, opened 2021-03-15
Deepchem Tutorial 14 issue


SaharRohaniZ:
## 🐛 Bug

I ran Tutorial 14 ( Modeling Protein-Ligand Interactions with Atomic Convolutions) but face an error with load_pdbbind()
## To Reproduce

1.Run all the steps in Tutorial 14

## Additional context

![featurizer error](https://user-images.githubusercontent.com/4704553/111096153-24dcb080-84fc-11eb-9791-3614707d0092.PNG)

ncfrey:
Hi @SaharRohaniZ, I just saw this bug as well - it was introduced because of a change to the `ComplexFeaturizer` parent class. I am working on a fix now and I'll let you know when it's up!

ncfrey:
Should be closed by #2441.

PR #2441, opened 2021-03-15
complex featurizer output fix

ncfrey:
## Description

There was an errant `}` that got left into a line in Tutorial 14. Also the `RdkitGridFeaturizer` failed featurization handling caused an error for `AtomicConvFeaturizer`, so I've fixed that. I wanted to do this quickly because of the 2.5.0 release, but there should probably be more robust handling for failured complex featurizations because the output for complex featurizers is not standard. This addresses #2438 

## Type of change

CCing @rbharath


rbharath (approved):
Sorry for the slow review! Looks good :)

PR #2463, opened 2021-04-01
Fixed error in computing Pearson correlation coefficient

peastman:
Previously we were directly using the SciPy implementation of `pearsonr()` as our metric.  That function actually returns a tuple, `(r, p_value)`.  The Metric class would then attempt to average over tasks, and so return the average of those two values.  If the inputs were perfectly correlated, it would return 0.5 instead of 1!


peastman:
The test failures are just from `test_mol2vec_fingerprint()`, which is currently failing on the main branch.

rbharath (approved):
Thanks for the fix! Feel free to merge when ready :)

PR #2565, opened 2021-06-23
Update FASTA Loader to accept arbitrary featurizers

alat-rights:
# Update FASTA Loader to accept arbitrary featurizers

## Description

Currently, the FASTALoader class only uses one hot encoding to featurize FASTA files.

It would be useful if the loader could be agnostic to featurizer, and allow users to use whatever featurizer they'd like.

This change is currently breaking, but I hope to make it non-breaking by making the default featurizer the one hot featurizer and thereby allowing one hot featurization by default.

- [X] Working with one hot featurizer
  - [X] Migrating from Biopython tools to DeepChem one hot featurizer
  - [X] Tests passing for one hot featurization
- [ ] Working with protein tokenizer 
  - [X] Making FASTALoader read FASTA files into numpy arrays of strings
  - [ ] Testing and iteration to ensure compatibility (on hold as protein tokenizer is not complete)
- [X] Check OneHotFeaturizer tests still passing
- [ ] ~~Working with arbitrary featurizers~~ (Currently not aware of any good candidate featurizers except for ProteinTokenizer which is still WIP)
  - [x] ~~New unit tests~~
- [ ] FASTA string validation (check that strings are FASTA format)
- [x] Test on real-world FASTA files (blocker on full-sized files: sharding support not yet implemented)

## Type of change

## Checklist

  - [X] Run `mypy -p deepchem` ~~and check no errors~~
    - **THREE ERRORS RELATED TO INCOMPATIBLE TYPES APPEAR TO BE  ERRONEOUS** *

\* Mypy errors:
```
deepchem/feat/molecule_featurizers/one_hot_featurizer.py:54: error: Incompatible types in assignment (expression has type "None", variable has type "int")
deepchem/data/data_loader.py:933: error: Incompatible types in assignment (expression has type "None", variable has type "Tuple[str, ...]")
deepchem/data/data_loader.py:937: error: Incompatible types in assignment (expression has type "None", variable has type "Tuple[str, ...]")
```


deepchem/data/data_loader.py
```diff
@@ -875,9 +875,88 @@ class provides convenience files to lead FASTA data and
   learning tasks.
   """
 
-  def __init__(self):
-    """Initialize loader."""
-    pass
+  def __init__(self,
+               featurizer=OneHotFeaturizer,
```
rbharath, line 879: You're passing in a class rather than a featurizer object here :). I think you want to do something like
```
featurizer=OneHotFeaturizer()
```
to make an object. I think that should work. Also make sure to add a type signature for `featurizer`

rbharath: Ok just noticed below that you initialize `OneHotFeaturizer` with different arguments for different charsets. A better default might be to use
```
featurizer=None
```
and below have the featurizer instantiated explicitly as a `OneHotFeaturizer` if `None` is passed in via the constructor.

deepchem/data/data_loader.py
```diff
@@ -875,9 +875,88 @@ class provides convenience files to lead FASTA data and
...
+
+    # Set user defined featurizer
+    self.user_specified_features = None
+    if isinstance(featurizer, UserDefinedFeaturizer):
+      self.user_specified_features = featurizer.feature_fields
+
+    # Initialize featurizer
+    try:
+      if self.charset is not None and self.max_length != -1:
+        featurizer = featurizer(
```
rbharath, line 948: You should have a check `if featurizer == None` and explicitly just use `OneHotFeaturizer` here

deepchem/data/data_loader.py
```diff
@@ -875,9 +875,88 @@ class provides convenience files to lead FASTA data and
...
+            charset=self.charset, max_length=self.max_length)
+      elif self.charset is not None and self.max_length == -1:
+        featurizer = featurizer(charset=self.charset)
+      elif self.charset is None and self.max_length != -1:
+        featurizer = featurizer(max_length=self.max_length)
+      elif self.charset is None and self.max_length == -1:
+        featurizer = featurizer()
+    except:
+      logger.exception("Sorry! Your featurizer may not be supported yet.")
```
rbharath, line 957: When we swap to accepting featurizer objects this exception won't make sense

deepchem/data/data_loader.py
```diff
@@ -908,11 +987,54 @@ def create_dataset(self,
...
+            if line[-1:] == '\n':  # Check last character in string
+              line = line[0:-1]  # Remove last character
+            sequence.append(line)
+        sequences = _add_sequence(sequences, sequence)
+        return sequences
+
+      def _add_sequence(sequences: np.array, sequence: list) -> np.array:
+        # Handle empty sequence
+        if sequence is None or len(sequence) <= 0:
+          logger.warning(
```
rbharath, line 1026: Could this spam many warnings? We want to make sure user logs won't be flooded with excess warnings if we expect this case to happen frequently

deepchem/data/tests/test_fasta_loader.py
```diff
@@ -16,7 +16,9 @@ def setUp(self):
     super(TestFASTALoader, self).setUp()
     self.current_dir = os.path.dirname(os.path.abspath(__file__))
 
-  def test_fasta_load(self):
+  def test_fasta_one_hot(self):
+    legacy = False  # Whether to assume legacy one hot encoding shape from FASTA loader.
```
rbharath, line 20: If something is legacy, we should deprecate it explicitly. Have we changed the expected shapes?

deepchem/feat/molecule_featurizers/one_hot_featurizer.py
```diff
@@ -32,21 +32,26 @@ class OneHotFeaturizer(Featurizer):
...
-      shorter than max_length, the SMILES is padded using space.
+    max_length: Optional[int], optional (default 100)
+      The max length for string. If the length of string is shorter than
+      max_length, the string is padded using space.
     """
     if len(charset) != len(set(charset)):
       raise ValueError("All values in charset must be unique.")
     self.charset = charset
-    self.max_length = max_length
+    if max_length is not None:
```
rbharath, line 51: Why would we want `self.max_length` to be None? Does that specify arbitrary length strings are allowed? If so, we should document that above in the docstring

deepchem/feat/molecule_featurizers/one_hot_featurizer.py
```diff
@@ -151,7 +156,11 @@ def pad_string(self, string: str) -> str:
     str
       String space padded to self.pad_length
     """
-    return string.ljust(self.max_length)
+    if isinstance(self.max_length, int):
+      return string.ljust(self.max_length)
+    else:
+      logger.info("DID NOT PAD: max_length is not int.")
```
rbharath, line 162: This warning could spam a lot of logging messages. There's a balance to strike in that we don't want to flood user logs with messages while also returning useful information

rbharath (commented):
Did a first review pass. The core structure looks good but we need to clean up a few things with the type handling.

deepchem/data/tests/test_fasta_loader.py
alat-rights: Yes, the shapes changed.

deepchem/feat/molecule_featurizers/one_hot_featurizer.py
alat-rights: Yes, it specifies arbitrary length strings allowed.

That makes sense.

deepchem/feat/molecule_featurizers/one_hot_featurizer.py
alat-rights: Sounds good.

deepchem/data/data_loader.py
alat-rights: I decided to go for a slightly different design where I just instantiate it with no parameters and then adjust the attributes after.

Is that acceptable?

deepchem/data/data_loader.py
alat-rights: image
It now looks like this.

alat-rights: Maybe this design is shortsighted and I should just have the user pass in something like OneHotFeaturizer(charset, max_length) instead?

deepchem/data/data_loader.py
alat-rights: For now, I changed it into logger.info.

I think it's generally quite unexpected for the FASTA file to be filled with empty sequences.

alat-rights:
Just added support for a legacy toggle and resolved lots of comments that were made earlier during meetings!

deepchem/data/data_loader.py
```diff
@@ -875,9 +876,71 @@ class provides convenience files to lead FASTA data and
...
+    legacy: bool (default True)
+      Whether to use legacy logic for featurization. Legacy mode will create
+      a one hot encoding of the FASTA content of shape
+      (number of FASTA sequences, number of channels + 1, max length, 1).
+
+      Legacy mode is only tested for ACTGN charsets, and will be deprecated.
+   """
+
+    # Process legacy toggle
+    if legacy:
```
rbharath, line 913: For deprecations, you want to use `warnings`. See https://github.com/deepchem/deepchem/blob/master/deepchem/models/graph_models.py#L1216 for a style example

deepchem/data/data_loader.py
```diff
@@ -875,9 +876,71 @@ class provides convenience files to lead FASTA data and
...
+      logger.info("""
+                  Deprecation warning: Legacy mode will soon be deprecated.
+                  Disable legacy mode by passing legacy=False during
+                  construction of FASTALoader object.
+                  """)
+      if featurizer is not None or auto_add_annotations:
+        logger.warning(f"""
```
rbharath, line 920: I'd say just raise a value error here instead of trying to correct. Usually better to fail fast rather than proceed since users don't notice warnings often

deepchem/data/data_loader.py
```diff
@@ -907,13 +970,64 @@ def create_dataset(self,
...
+              line = line[0:-1]  # Remove last character
+            sequence = np.append(sequence, line)
+        sequences = _add_sequence(sequences, sequence)
+        return sequences
+
+      def _add_sequence(sequences: np.array, sequence: np.array) -> np.array:
+        # Handle empty sequence
+        if sequence is None or len(sequence) <= 0:
+          # TODO log attempts to add empty sequences every shard
+          """
```
rbharath, line 1015: Can you remove the old code here?

deepchem/feat/molecule_featurizers/one_hot_featurizer.py
```diff
@@ -88,14 +96,14 @@ def _featurize_string(self, string: str) -> np.ndarray:
...
       The index of unknown character is `len(charset)`.
     """
-    # validation
-    if (len(string) > self.max_length):
-      logger.info(
-          "The length of {} is longer than `max_length`. So we return an empty array."
-      )
-      return np.array([])
+    if self.max_length is not None:
+      if (len(string) > self.max_length):  # Validation
```
rbharath, line 100: I'd say here we should probably just fail with a ValueError instead of trying to recover

rbharath (commented):
A few more comments on how to handle warnings/deprecations properly. Should be good in one more round I think.

rbharath:
Ok test failures look to be unrelated. Going to merge this in!

PR #2577, opened 2021-07-01
[WIP] Weights & Biases Logger (Standardized logger API)

kshen3778:
# Pull Request Template

## Description

Refactored logging code to use the new structure described in https://github.com/deepchem/deepchem/issues/2580
- Addition of new standardized methods used for Loggers

Expanded KerasModel and TorchModel model checkpointing functionality to support saving to W&B Artifacts
- Added `save_checkpoint` method in `WandbLogger` to handle checkpointing to Wandb.

Status: Work in progress

## Type of change

## Checklist


deepchem/models/wandblogger.py
```diff
@@ -126,24 +140,76 @@ def log_data(self,
...
+          model_name = checkpoint_name + "_" + self.wandb_run.name + ".pt"
+          model_path = os.path.abspath(os.path.join(path, model_name))
+          torch.save(data, model_path)
+
+          # Upload Artifact
+          artifact = self._wandb.Artifact(model_name,
+                                          type='model',
+                                          metadata=metadata)
+          artifact.add_file(model_path)
+          self.wandb_run.log_artifact(artifact, aliases=["latest", value_name + "=" + str(value)])
```
morganmcg1, line 215: add "step=step" to alias

kshen3778:
@peastman Could you do a quick pass to see if this is in line with what we were [discussing]?(https://github.com/deepchem/deepchem/issues/2580)? I haven't added styling or docstrings yet.

I noticed in the tests that because WandbLogger now requires importing both tensorflow and torch, it would throw `ModuleNotFoundError` in the tests. Deepchem Tensorflow Tests cannot import torch and vice versa (assuming I'm reading the logs correctly). Any suggestions on how to handle this or if there's a preferred way?

deepchem/models/__init__.py
```diff
@@ -4,10 +4,17 @@
 # flake8: noqa
 from deepchem.models.models import Model
 from deepchem.models.multitask import SingletaskToMultitask
-from deepchem.models.wandblogger import WandbLogger
 from deepchem.models.callbacks import ValidationCallback
 
-# Tensorflow Depedency Models
+from deepchem.models.logger import Logger
```
peastman, line 9: I don't think you check this file in.

deepchem/models/callbacks.py
```diff
@@ -2,6 +2,7 @@
 Callback functions that can be invoked while fitting a KerasModel.
 """
 import sys
+from deepchem.models.wandblogger import WandbLogger
```
peastman, line 5: ValidationCallback shouldn't have any explicit references to WandbLogger.  The idea of having a generic Logger class is that all common code (like ValidationCallback) will reference it through the generic API without needing to know exactly what kind of Logger it is.

deepchem/models/keras_model.py
```diff
@@ -186,28 +188,41 @@ def __init__(self,
...
     if wandb and not _has_wandb:
-      logger.warning(
+      logs.warning(
           "You set wandb to True but W&B is not installed. To use wandb logging, "
-          "run `pip install wandb; wandb login`")
+          "run `pip install wandb` then log in using `wandb login` "
+          "or `import wandb; wandb.login()` in script.")
     self.wandb = wandb and _has_wandb
 
+    # Wandb_logger flag support (DEPRECATED)
```
peastman, line 212: `wandb_logger` should be completely removed, replaced by `logger`.  It hasn't existed in any release, so we don't need to worry about maintaining backward compatibility.

deepchem/models/keras_model.py
```diff
@@ -460,28 +476,49 @@ def fit_generator(self,
 
       if checkpoint_interval > 0 and current_step % checkpoint_interval == checkpoint_interval - 1:
         manager.save()
+        for ext_logger in self.loggers:
+          if isinstance(ext_logger, WandbLogger):
+            ext_logger.save_checkpoint(self.model_dir,
```
peastman, line 481: It's up to the logger to decide when to write checkpoints.  `checkpoint_interval` tells the model when it should write checkpoints using the standard mechanism.  If a logger is writing checkpoints with a different mechanism, that isn't the model's concern.

deepchem/models/keras_model.py
```diff
@@ -460,28 +476,49 @@ def fit_generator(self,
...
       if self.tensorboard and should_log:
         self._log_scalar_to_tensorboard('loss', batch_loss, current_step)
-      if (self.wandb_logger is not None) and should_log:
-        all_data = dict({'train/loss': batch_loss})
-        self.wandb_logger.log_data(all_data, step=current_step)
+      for ext_logger in self.loggers:
+        if isinstance(ext_logger, WandbLogger):
+          ext_logger.log_batch({"loss": batch_loss}, current_step, inputs, labels, group="train")
+        else:
+          ext_logger.log_batch({"loss": batch_loss}, current_step, inputs, labels)
```
peastman, line 496: Everything should be done through the generic interface.  The whole point of a generic interface is that you don't need special case code for different subclasses.

deepchem/models/keras_model.py
```diff
@@ -460,28 +476,49 @@ def fit_generator(self,
...
       manager.save()
+      for ext_logger in self.loggers:
+        if isinstance(ext_logger, WandbLogger):
+          ext_logger.save_checkpoint(self.model_dir,
+                                     self,
+                                     "train_checkpoints",
+                                     "step",
+                                     current_step,
+                                     max_checkpoints_to_keep,
+                                     checkpoint_on_min=False)
```
peastman, line 518: The above issues apply here too.  1) There shouldn't be special handling for WandbLogger, and 2) checkpointing by the logger isn't the model's concern.  Logger should define a generic method that gets called at the end of `fit()`.  A subclass can do whatever it wants in that method: save a checkpoint, clean up resources, whatever.

deepchem/models/logger.py
```diff
@@ -0,0 +1,2 @@
+class Logger(object):
+    pass
```
peastman, line 2: Logger needs to provide all the methods that subclasses implement.  It defines the API, and subclasses provide implementations.  In the parent class, all the methods should just throw NotImplementedError.

deepchem/models/keras_model.py
kshen3778: I was under the assumption that the Logger would use the same checkpoint_interval and strategy as the main model class. If that's not the case then we can have each logger will define its own checkpointing mechanism. Which means for WandbLogger we can initialize it with it's own checkpoint_interval argument. Is my understanding correct?

deepchem/models/keras_model.py
kshen3778: Are you suggesting something like this:
```
for ext_logger in self.loggers:
    ext_logger.log_batch({"loss": batch_loss}, current_step, inputs, labels, group="train")
```
If we don't check the type of the logger wouldn't it be problematic if the `log_batch` function signature differs between loggers? For example, the `log_batch` in WandbLogger adds an extra `group` parameter, but this parameter would just be a Wandb thing and not universal to all Loggers. The core log_batch function would be `def log_batch(data, current_step, inputs, labels)`.

deepchem/models/logger.py
kshen3778: Yup, will fill this in!

I'll add these:
```
log_batch
log_epoch
log_values
setup
finish
```
Not sure about `save_checkpoint()` , that might be something that individual loggers can define?

deepchem/models/__init__.py
kshen3778: True, will remove it.

deepchem/models/callbacks.py
kshen3778: I'm importing this since I make a check for the logger type later on in the code.
This actually is something you touch on in another comment. We can continue the discussion down below:
https://github.com/deepchem/deepchem/pull/2577#discussion_r676854073

deepchem/models/keras_model.py
kshen3778: Sure, I can remove it. This is concerning existing in major releases right? As far as I know, it's currently still possible for people to use `wandb_logger` if they install deepchem using `pip install --pre deepchem`

deepchem/models/keras_model.py
peastman: Correct!

deepchem/models/keras_model.py
peastman: All loggers will have identical signatures for all methods.  That's the point of a common interface.  The parent class defines the API, and all subclasses have to use it.

It's ok to include `kwargs` as part of the signature.  But if KerasModel and TorchModel will be passing an argument, it probably ought to be part of the standard API.

deepchem/models/keras_model.py
peastman: All the more reason to remove it now so we don't get stuck with another deprecated argument we have to support in the future.

kshen3778: sure, I've removed it

kshen3778:
@peastman Made some edits based on our conversation. I've modified the logger base class to include all the methods. In addition, the logger calls (log_batch, log_values) in the Model classes and ValidationCallback are all standardized now. Checkpointing has been moved into the logger classes and WandbLogger will use its own mechanism.

Haven't done docstrings yet, will do that at the end when the code is finalized.

deepchem/models/logger.py
```diff
@@ -0,0 +1,64 @@
...
+  def log_batch(self,
+                loss: Dict,
+                step: int,
+                inputs: tensor,
+                labels: tensor,
+                location: Optional[str] = None,
+                model: Optional[Model] = None,
+                checkpoint_metric: Optional[str] = None,
+                checkpoint_metric_value: Optional[numeric] = None,
+                checkpoint_on_min: bool = True):
```
peastman, line 34: I've been debating about this.  I don't like this API too much.  I know I'm the one who suggested the logger's checkpointing mechanism should be independent of the standard one, but seeing what it does to the API I'm reconsidering it.  This forces each logger to repeat a bunch of logic that really is specific to ValidationCallback and unrelated to logging.  It might be better to just make `save_checkpoint()` public and call it whenever we write a checkpoint normally.  What do you think?

deepchem/models/keras_model.py
```diff
@@ -186,28 +185,36 @@ def __init__(self,
       self.optimizer = optimizer
     self.tensorboard = tensorboard
 
+    # Create a list of loggers
+    self.loggers = list()
+    if logger is not None:
+      if isinstance(logger, list):
```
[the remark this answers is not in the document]
kshen3778: Will do!

deepchem/models/logger.py
```diff
@@ -0,0 +1,64 @@
...
+
+  def finish(self):
+    raise NotImplementedError
+
+  def log_batch(self,
+                loss: Dict,
+                step: int,
+                inputs: tensor,
+                labels: tensor,
+                location: Optional[str] = None,
```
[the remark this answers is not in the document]
kshen3778: This is basically the `group` parameter from before. It's a string to uniquely identify where the logging is happening so when viewed in Wandb dashboard, the metric charts are under the correct headings (train/, callback1/, callback2/, etc).  Otherwise, you could potentially have a train and a callback both logging the same metric (for example loss) and the loggers wouldn't be able to distinguish. I figured renaming it to `location` sounds better since "group" isn't really meaningful.

Originally I was going to have the "location" param just for the WandbLogger but since we agreed that the API should be common and have the same parameters, I added it to the generic Logger API and made it optional.

deepchem/models/keras_model.py
```diff
@@ -216,14 +223,15 @@ def __init__(self,
...
         tensorboard=tensorboard,
         log_frequency=log_frequency)
-    wandb_logger_config.update(**kwargs)
+    logger_config.update(**kwargs)
 
-    if self.wandb_logger is not None:
-      self.wandb_logger.update_config(wandb_logger_config)
+    # Setup and initialize external loggers
+    for ext_logger in self.loggers:
+      ext_logger.setup(logger_config)
```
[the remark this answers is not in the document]
kshen3778: My bad, I might've misunderstood the previous suggestion as passing in just a config versus passing in the model. So something like this essentially:
```
for ext_logger in self.loggers:
  ext_logger.setup(self) # where "self" is the Torch/KerasModel
```
and then WandbLogger's setup():
```
def setup(self, model):
    self.model = model
    if self._wandb.run is None:
      self.wandb_run = self._wandb.init(**self.wandb_init_params)
      if self.wandb_run is not None:
        self.wandb_run.config.update(get config from model) # add model config to wandb config
    else:
      self.wandb_run = self._wandb.run
    self.initialized = True
```
Is my understanding correct? If that's the case then I think it makes sense and would be my approach too.

deepchem/models/logger.py
kshen3778: I wasn't sure at the beginning either, but after working through both approaches I think having the integrated checkpointing (our original approach, the one you described above) seems to be a bit more clean from a code perspective. If we keep the independent checkpointing approach (where save_checkpoint is called in log_batch, log_values) we would have to pass in a lot of checkpointing parameters as you probably saw above (checkpoint_metric, checkpoint_metric_value, checkpoint_on_min) to the logging functions. 

We could expose `save_checkpoint()` as a Logger API method and it would be called anywhere whereever checkpointing is done. The three places I've identified where checkpointing is called are:
1. Model class fit_generator(): https://github.com/kshen3778/deepchem/blob/wandb2/deepchem/models/keras_model.py#L469
2. Model class end of fit: https://github.com/kshen3778/deepchem/blob/wandb2/deepchem/models/keras_model.py#L498
3. ValidationCallback: https://github.com/kshen3778/deepchem/blob/wandb2/deepchem/models/callbacks.py#L96
Let me know if I'm missing any. 
So I'm guessing is that in each of these locations there would be:
```
for ext_logger in self.loggers:
   ext_logger.save_checkpoint(.....)
```
I'm still unsure what the parameters for `save_checkpoint()` would be. I'll add another post once I give it some more thought.

I also don't want to touch the [pre-existing](https://github.com/kshen3778/deepchem/blob/wandb2/deepchem/models/keras_model.py#L1072) save_checkpoint() that exists in the Model classes because they are specific to the default Tensorflow/Pytorch checkpointing behaviour.


On a sort of related note, the one annoying challenge I came across was the checkpoint naming. We want checkpoints coming from different locations (training, eval callback 1, eval callback 2, etc) to have different names when being saved. Right now in Wandb, for each "location" we have all of the location's checkpoints' being saved under `[location].[metric-name].checkpoints_[run_name]`. 

Screenshot from 2021-08-05 09-59-57
Screenshot from 2021-08-05 10-01-19

This is one reason I added a `name` parameter to ValidationCallback in order to distinguish between different callbacks. The name would be [passed into](https://github.com/kshen3778/deepchem/blob/wandb2/deepchem/models/callbacks.py#L110) `location` parameter when logging (and saving checkpoints). The annoying part comes when the user doesn't provide a name for the ValidationCallback in which case we have to [assign it a unique name](https://github.com/kshen3778/deepchem/blob/wandb2/deepchem/models/callbacks.py#L102). I feel like this isn't ideal but I'm not sure if there's a better way. In theory we could name checkpoints under their model_dir but that wouldn't work if mutliple callbacks are sharing a save directory.

kshen3778: @peastman I've written up my thoughts above. Anything else you want to add for this?

peastman: I think I agree.  Let's just make `save_checkpoint()` public and call it anywhere that checkpoints get written.

The "location" argument seems a bit too implementation specific, relating to how it's being saved.  Am I correct that this really just refers to what information is being saved: training loss, validation loss (perhaps for more than one validation set, though that's unusual), etc.?  The ideal would be to pass in semantically meaningful information, and have WandbLogger select a location based on that.

kshen3778: Sounds good for `save_checkpoint()`

Regarding the "location" argument: It doesn't necessarily refers to the information being saved. It's more related to the organization system used by the logger dashboard. See this [run](https://wandb.ai/kwshen/deepchem/runs/29oukj1f?workspace=user-kwshen) for example:
Screenshot from 2021-08-16 15-30-20
Screenshot from 2021-08-16 15-30-31
Basically the locations in this case are eval/callback1, eval/callback2, and train. This is because there are two ValidationCallbacks running as well as the training (the two callbacks might be operating on different datasets - for example "val" and "test" or two validation sets).  This structure allows people to understand exactly where their values are coming from. Otherwise, you could very possibly have pearson_r2_score being logged in multiple places (for example in two callbacks, or once in training and once in one the callbacks) and you wouldn't be able to distinguish where it's coming from.

I think it's not just W&B that has this hierarchical location structure. I saw that Neptune, for example also uses this style (https://docs.neptune.ai/you-should-know/logging-metadata). I'm sure other loggers probably follow similar formats.

kshen3778:
@peastman I've reverted back to the original format where `save_checkpoint` is a logger method. Please take a look and let me know if this structure makes more sense and if I'm missing anything. I can also remove the old `wandb` boolean flag in Keras/TorchModel if you'd like.

deepchem/models/keras_model.py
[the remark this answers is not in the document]
peastman: Don't forget to make this change.

deepchem/models/callbacks.py
```diff
@@ -93,7 +95,24 @@ def __call__(self, model, step):
...
+          ext_logger.save_checkpoint(
+              self.save_dir,
+              model,
+              checkpoint_name,
+              self.metrics[self.save_metric].name,
+              checkpoint_score,
+              max_checkpoints_to_track=5,
+              checkpoint_on_min=self.save_on_minimum)
+    for ext_logger in model.loggers:
+      ext_logger.log_values(scores, step, location="eval/" + self.name)
```
peastman, line 118: What about giving it a more informative name, for example `location=f'validation-{self.name}/'`?

deepchem/models/keras_model.py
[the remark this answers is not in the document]
peastman: Exactly.

deepchem/models/logger.py
```diff
@@ -0,0 +1,142 @@
...
+    raise NotImplementedError
+
+  def save_checkpoint(self,
+                      model_dir: str,
+                      dc_model: Model,
+                      checkpoint_name: str,
+                      value_name: str,
+                      value: numeric,
+                      max_checkpoints_to_track: int,
+                      checkpoint_on_min: bool,
```
peastman, line 116: I think these arguments are all unnecessary now.

deepchem/models/logger.py
```diff
@@ -0,0 +1,142 @@
...
+    Parameters
+    ----------
+    data: Dict
+      Training summary values to be logged
+    """
+    raise NotImplementedError
+
+  def save_checkpoint(self,
+                      model_dir: str,
+                      dc_model: Model,
```
peastman, line 111: These two arguments shouldn't be needed.  The logger already got a reference to the Model in `setup()`.  It can save it there if it will need it later.  And it can get the model directory from the model.

deepchem/models/wandblogger.py
```diff
@@ -78,67 +86,224 @@ def __init__(self,
...
     """
-    self.wandb_run.config.update(config_data)
+
+    # Only called once when first checkpoint is saved to create tracking record
+    if (checkpoint_name not in self.best_models):
+      # Set up to track top values of this checkpoint
+      self.best_models[checkpoint_name] = {}
+      self.best_models[checkpoint_name]["model_values"] = []
+
+    # Sort in order
```
peastman, line 244: All of the following is repeating a lot of logic that really belongs in ValidationCallback.  `save_checkpoint()` should simply save a checkpoint and nothing more.

deepchem/models/keras_model.py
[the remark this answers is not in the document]
kshen3778: Didn't see this sorry. Made the change

deepchem/models/callbacks.py
kshen3778: Sure, I've renamed it.

deepchem/models/logger.py
kshen3778: Removed.

deepchem/models/logger.py
kshen3778: Removed

deepchem/models/wandblogger.py
kshen3778: Yup agreed, I realized ValidationCallback already had logic for when to save checkpoints. I've removed the logic from save_checkpoint()

kshen3778:
Just a note: Import checking in "Test for DeepChem Tensorflow" seem to be failing: `python -c "import deepchem; import tensorflow;"`
Not sure if this is related to the Logger changes made, would be good to have someone check just in case.

rbharath:

PR #2586, opened 2021-07-07
WandbLogger fixes: removed finish() from Model fit(), ValidationCallbacks fixes

kshen3778:
# Pull Request Template

## Description

This PR fixes these existing problems with WandbLogger:
- [x] Removed finish() from KerasModel.fit() and TorchModel.fit()

Previously, fit() completed ended the wandb run, which meant users could not call fit() multiple times (wandb process is shutdown)

- [x] Allow multiple ValidationCallbacks to be used for logging (keys logged to Wandb differentiate between different VC) 

Previously, if there were two or more ValidationCallbacks, Wandb would not be able to distinguish them when logging especially if they are on different datasets. If these VCs are logging the same metrics, there will still be only one chart in the Wandb dashboard.
Now it is possible to have multiple VCs each scoring with a different dataset. These metrics (keys) will be a given a unique when logged.
```
vc_train = dc.models.ValidationCallback(train_dataset, 1, [metric])
vc_valid = dc.models.ValidationCallback(valid_dataset, 1, [metric])
model.fit(train_dataset, nb_epoch=10, callbacks=[vc_train, vc_valid])
```
Note: In the future if support for logging multiple datasets in ValidationCallback is added, this modification/fix can be removed. 

- [x] ValidationCallback wandblogger existance check: https://github.com/deepchem/deepchem/pull/2586#discussion_r667101882

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/callbacks.py
+++ b/deepchem/models/callbacks.py
@@ -90,4 +90,4 @@ class ValidationCallback(object):
     if model.wandb or (model.wandb_logger is not None):
       # Log data to Wandb
       data = {'eval/' + k: v for k, v in scores.items()}
-      model.wandb_logger.log_data(data, step)
+      model.wandb_logger.log_data(data, step, dataset_id=id(self.dataset))
--- a/deepchem/models/callbacks.py
+++ b/deepchem/models/callbacks.py
@@ -87,7 +87,7 @@ class ValidationCallback(object):
       if self._best_score is None or score < self._best_score:
         model.save_checkpoint(model_dir=self.save_dir)
         self._best_score = score
-    if model.wandb or (model.wandb_logger is not None):
+    if model.wandb_logger is not None:
       # Log data to Wandb
       data = {'eval/' + k: v for k, v in scores.items()}
       model.wandb_logger.log_data(data, step, dataset_id=id(self.dataset))
--- a/deepchem/models/tests/test_kerasmodel.py
+++ b/deepchem/models/tests/test_kerasmodel.py
@@ -319,15 +319,19 @@ def test_wandblogger():
        tf.keras.layers.Dense(1)])
   model = dc.models.KerasModel(
       keras_model, dc.models.losses.L2Loss(), wandb_logger=wandblogger)
-  vc = dc.models.ValidationCallback(valid_dataset, 1, [metric])
-  model.fit(train_dataset, nb_epoch=10, callbacks=[vc])
+  vc_train = dc.models.ValidationCallback(train_dataset, 1, [metric])
+  vc_valid = dc.models.ValidationCallback(valid_dataset, 1, [metric])
+  model.fit(train_dataset, nb_epoch=10, callbacks=[vc_train, vc_valid])
+  # call model.fit again to test multiple fit() calls
+  model.fit(train_dataset, nb_epoch=10, callbacks=[vc_train, vc_valid])
+  wandblogger.finish()
 
   run_data = wandblogger.run_history
   valid_score = model.evaluate(valid_dataset, [metric], transformers)
 
   assert math.isclose(
       valid_score["pearson_r2_score"],
-      run_data['eval/pearson_r2_score'],
+      run_data['eval/pearson_r2_score_(1)'],
       abs_tol=0.0005)
 
 
--- a/deepchem/models/wandblogger.py
+++ b/deepchem/models/wandblogger.py
@@ -1,6 +1,6 @@
 import logging
 import importlib.util
-from typing import Optional, Union
+from typing import Optional, Union, List
 
 logger = logging.getLogger(__name__)
 
@@ -83,6 +83,9 @@ class WandbLogger(object):
     self.wandb_init_params.update(**kwargs)
     self.initialized = False
 
+    # Dataset ids are used to differentiate datasets seen by the logger
+    self.dataset_ids = [] # type: List[Union[int, str]]
+
   def setup(self):
     """Initializes a W&B run and create a run object.
     If a pre-existing run is already initialized, use that instead.
@@ -93,7 +96,7 @@ class WandbLogger(object):
       self.wandb_run = self._wandb.run
     self.initialized = True
 
-  def log_data(self, data, step):
+  def log_data(self, data, step, dataset_id: Optional[Union[int, str]] = None):
     """Log data to W&B.
 
     Parameters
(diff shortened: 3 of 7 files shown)


rbharath (commented):
I did a quick pass and this looks good, but not an expert at the logging code :). 

@peastman Could you review this PR?

deepchem/models/wandblogger.py
```diff
@@ -83,6 +83,9 @@ def __init__(self,
     self.wandb_init_params.update(**kwargs)
     self.initialized = False
 
+    # Dataset ids are used to differentiate datasets seen by the logger
+    self.dataset_ids = []  # type: List[Union[int, str]]
```
peastman, line 87: It would be better to use a type annotation rather than a comment:

```python
self.dataset_ids: List[Union[int, str]] = []
```

deepchem/models/tests/test_kerasmodel.py
```diff
@@ -319,15 +319,19 @@ def test_wandblogger():
...
   model = dc.models.KerasModel(
       keras_model, dc.models.losses.L2Loss(), wandb_logger=wandblogger)
-  vc = dc.models.ValidationCallback(valid_dataset, 1, [metric])
-  model.fit(train_dataset, nb_epoch=10, callbacks=[vc])
+  vc_train = dc.models.ValidationCallback(train_dataset, 1, [metric])
+  vc_valid = dc.models.ValidationCallback(valid_dataset, 1, [metric])
+  model.fit(train_dataset, nb_epoch=10, callbacks=[vc_train, vc_valid])
+  # call model.fit again to test multiple fit() calls
+  model.fit(train_dataset, nb_epoch=10, callbacks=[vc_train, vc_valid])
+  wandblogger.finish()
```
peastman, line 327: Will users need to call `finish()`, or are you only doing this so you can examine the log?  If someone just trains a model normally, is it ok for them to skip calling it?

deepchem/models/callbacks.py
```diff
@@ -90,4 +90,4 @@ def __call__(self, model, step):
     if model.wandb or (model.wandb_logger is not None):
```
peastman, line 90: It not directly related to this PR, but shouldn't this line be just

```python
if model.wandb_logger is not None:
```

You reference `model.wandb_logger` inside the block, so you'll get an exception if it's `None`.

deepchem/models/wandblogger.py
```diff
@@ -93,7 +96,7 @@ def setup(self):
       self.wandb_run = self._wandb.run
     self.initialized = True
 
-  def log_data(self, data, step):
+  def log_data(self, data, step, dataset_id: Optional[Union[int, str]] = None):
```
peastman, line 99: Could you add type annotations for the other two arguments as well?

deepchem/models/tests/test_kerasmodel.py
kshen3778: Normally, it's best practice to call `finish()` for wandb since that ends the run and shuts off the wandb backend process. Without calling finish, nothing will crash but the wandb process will still be running until the next run or once the notebook/script is closed/shutoff (it might also prevent files such as checkpoints from being uploaded).

Currently `wandblogger.finish()` also saves the run history data for testing usage. I can move the saving run history data into its own function if you'd like? But I think we would still need to call finish() at some point.

If you don't think users should call `wandblogger.finish()`, is there a better place where you think I should add finish() to end the run? fit() was the only place that I could think of originally.

deepchem/models/wandblogger.py
kshen3778: Will fix this!

deepchem/models/tests/test_kerasmodel.py
peastman: There isn't really any good place for it.  The typical usage is 1) create the model, 2) train it by calling `fit()` some number of times, 3) evaluate it either with `evaluate()` or in some other manner.  If the script exits without having called `finish()`, will the process get shut down and all data saved correctly?  If so, I think that's ok.

kshen3778: `wandblogger.finish()` makes a call to [wandb.finish()](https://docs.wandb.ai/ref/python/finish) at the end. However as it says on the doc `wandb.finish()` will also be called automatically when a script exits. I'm not too certain about the behaviour for notebooks since you could still run cells after the run finishes and I'm not sure if shutting down a notebook is equivalent to exiting a script. I always explicitly call finish() just to make sure all data is saved and uploaded.

So I suppose it would still be okay if users don't call `wandblogger.finish()` but I think it makes sense to mention it in the docs for best practices.

peastman: Is there something you can call that ensures all data is saved and uploaded, but doesn't shut down the process?

kshen3778: I went back and checked with the team on this: There would be no risk of data not being saved or uploaded if `finish()` is not called. `finish()` just ends the wandb python process (in a small minority of scenarios this might be important). A user could choose to not call `finish()` and they would be perfectly fine. Apologies for the confusion earlier.

Most users do call `wandb.finish()` and in other integrations with libraries such as Huggingface or Pytorch Lightning users might explicitly call `wandb.finish()` if there's no `finish()` wrapper function inside the Wandb logger class. They might even save additional data after training the model and then once they decide they are done they would call `wandb.finish()`. We also emphasize `wandb.finish()` a lot in our docs too.

Since our WandbLogger is importing wandb inside the class itself (importing wandb in the logger class is a standard we are gradually moving towards), we don't require users to import wandb at the beginning of their code and so there's no good way to call `wandb.finish()` since wandb is not always imported. Therefore, we would just expose a `WandbLogger.finish()` function that calls `wandb.finish()` so that users can shut down the wandb process.
Our opinion is that we think that `wandblogger.finish()` should be kept just so that it's clear that users can access something that calls wandb.finish() if they need to.

peastman: That sounds fine to me.

kshen3778:
@peastman I'm done with all changes - it's ready for merge now if everything else is good. Some of the tests seem to be cancelled so not sure if anything is wrong on that front

peastman:
Sorry for not replying sooner!  I don't think any of the test failures are related to this change.  I'll merge it.

kshen3778:
Perfect, thanks!

Issue #2564, opened 2021-06-22
Errors in SineCoulombMatrix


peastman:
I'm trying to fix some of the test failures that have appeared in the main branch.  I tracked the failure in `test_sine_coulomb_matrix()` down to this line:

https://github.com/deepchem/deepchem/blob/cbab9678fa8aa480996bcb030f7fedcd9950ca3d/deepchem/feat/material_featurizers/sine_coulomb_matrix.py#L80

It throws this exception:

```
  File "/Users/peastman/miniconda3/envs/dc/lib/python3.8/site-packages/matminer/featurizers/structure/__init__.py", line 22, in <module>
    from matminer.featurizers.structure.sites import SiteStatsFingerprint
  File "/Users/peastman/miniconda3/envs/dc/lib/python3.8/site-packages/matminer/featurizers/structure/sites.py", line 9, in <module>
    from matminer.featurizers.site import (
  File "/Users/peastman/miniconda3/envs/dc/lib/python3.8/site-packages/matminer/featurizers/site/__init__.py", line 9, in <module>
    from matminer.featurizers.site.fingerprint import (
  File "/Users/peastman/miniconda3/envs/dc/lib/python3.8/site-packages/matminer/featurizers/site/fingerprint.py", line 33, in <module>
    with open(os.path.join(os.path.dirname(__file__), "cn_target_motif_op.yaml"), "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/peastman/miniconda3/envs/dc/lib/python3.8/site-packages/matminer/featurizers/site/cn_target_motif_op.yaml'
```

I have matminer 0.7.2 installed, which is the most recent release.  This looks to me like matminer is just broken.

peastman:
I just reported it at https://matsci.org/t/error-importing-matminer/37113.  Let's see what they say.

peastman:
This error is fixed in matminer 0.7.3.

PR #2590, opened 2021-07-09
Remove pin for matminer

peastman:
The latest version of matminer supposedly fixes the problem in #2564.  Pinning to an old version should no longer be necessary.


rbharath (approved):
LGTM. Please feel free to merge once ready!

peastman:
Lots of test failures, but I don't see the particular ones that were related to matminer.

PR #2595, opened 2021-07-13
Improved handling of classification_handling_mode in Metric

peastman:
This makes the changes suggested in https://github.com/deepchem/deepchem/issues/2490#issuecomment-856322491.  In particular:

- `classification_handling_mode` now has different values to specify "pick a mode for me" versus "pass the values to the metric unchanged".
- `threshold_predictions()` always uses `argmax()` if a threshold is not specified.  That means it now works if the inputs are logits rather than probabilities.
- The documentation is improved to better describe what it really does.


rbharath (approved):
LGTM. Feel free to merge when ready!

Issue #2588, opened 2021-07-07
Additional Support in New CI


VIGNESHinZONE:
I have made changes in Github Actions, which provides four separate packages in `deepchem` for the different environments - 
1. `Deepchem`
2. `Deepchem[tensorflow]`
3. `Deepchem[torch]`
4. `DeepChem[jax]`

Changes have been made in this https://github.com/deepchem/deepchem/pull/2573 . Its a long PR, and it has become difficult to manage it with new errors showing upon new PR's being merged into `Master` branch and hence I was asked by @rbharath @ncfrey to document all the errors showing so that we could manage them more easily.

### Main Feature of PR 
1. Separate test suites for different environments using Pytest Markers - `tensorflow`, `torch`, `jax`. Along with that, had to make `import` changes accordingly
2. Made changes in Github Actions accordingly and have added scripts for installing these packages separately.
3. There four environments in CI - `common`(all those which are only `numpy` or `scikit-learn` dependent) , `tensorflow`, `torch`, `jax`

### 1. DeepChem Common - 
Most of these errors are interconnected, fixing one might solve the whole bucket

* Errors in `FastaLoader` - 

- [ ] `TestFASTALoader.test_fasta_one_hot`
- [ ] `TestFASTALoader.test_fasta_one_hot_big`
- [ ] `TestFASTALoader.test_legacy_fasta_one_hot`

* Errors of the form 
```
TypeError: join() argument must be str or bytes, not 'float64'
``` 

- [ ] `deepchem/data/tests/test_data_loader` - [error log](https://github.com/deepchem/deepchem/pull/2573/checks?check_run_id=3008679397#step:14:472)
- [ ] `deepchem/data/tests/test_datasets.py` - [error log](https://github.com/deepchem/deepchem/pull/2573/checks?check_run_id=3008679397#step:14:532)
- [ ] `deepchem/trans/tests/test_flattening.py` [error log](https://github.com/deepchem/deepchem/pull/2573/checks?check_run_id=3008679397#step:14:756)
- [ ] `deepchem/trans/tests/test_normalization.py` [error log](https://github.com/deepchem/deepchem/pull/2573/checks?check_run_id=3008679397#step:14:845)

* Errors in Gaussian HyperParameter - 
I was not able to reproduce the errors in my local environment or google collab [notebook](https://colab.research.google.com/drive/12BR3qFKBDKWrYkwUUPXcjBCTwBMVCw_9?usp=sharing)

[...]

PR #2605, opened 2021-07-16
Fixed errors caused by new Pandas version

peastman:
This fixes many of the errors in #2588.  They were caused by a change in behavior between Pandas 1.2 and 1.3.  When loading a DiskDataset, we load the metadata from a csv file and set any missing values to None:

https://github.com/deepchem/deepchem/blob/f3ba82cd015a7c59753baf4042092f1e01c56ea2/deepchem/data/datasets.py#L1220-L1221

In Pandas 1.3, if a column is completely empty, it sets the dtype for that column to float64 with every element being NaN.  Because None is not a legal value for a float64 column, the call to `where()` sets them to NaN instead.  This leads to errors later when we check for None values.

The fix is to tell it to use `object` as the dtype for all columns.

### DIFF
--- a/deepchem/data/datasets.py
+++ b/deepchem/data/datasets.py
@@ -1217,7 +1217,8 @@ class DiskDataset(Dataset):
       tasks_filename, metadata_filename = self._get_metadata_filename()
       with open(tasks_filename) as fin:
         tasks = json.load(fin)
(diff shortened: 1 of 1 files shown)


rbharath (approved):
Great find! This looks like a gnarly hard to find bug.

Feel free to merge when ready

PR #2645, opened 2021-08-05
Made a test less flaky

peastman:
This test was using too large a learning rate, which made it fail often.  Reducing it makes the test a lot more reliable.


rbharath (approved):
LGTM!

PR #2698, opened 2021-09-18
Docs ScaffoldSplitter

davidRFB:
# Pull Request Template

## Description
Addition of documentation in the ScaffoldSplitter class. Some basic definition and one example of the use. 

Fix #(issue)
https://github.com/deepchem/deepchem/issues/2688#issue-991296674 

## Type of change

## Checklist

### DIFF
--- a/deepchem/splits/splitters.py
+++ b/deepchem/splits/splitters.py
@@ -1315,6 +1315,35 @@ def _split_fingerprints(fps: List, size1: int,
 
 class ScaffoldSplitter(Splitter):
   """Class for doing data splits based on the scaffold of small molecules.
+  
+  The Bemis-Murcko scaffold representation in this splitter identifies rings,
+  linkers, frameworks ( combiantions between linkers and rings) and atomic properties 
+  such as atom type, hibridization and bond order in a dataset of molecules.
+
+
+  Examples
+  ---------
+  >>> import deepchem as dc
+  >>> # creation of demo data set with some smiles strings
+  ... data_test = ['CC(C)=CCCC(C)=CC(=O)', 'CCCC=C', 'CCCCCCCCCCCCCC', 'CC(C)Cl',
+  ...          'CCC(C)CO', 'CCCCCCCCCC(C)O', 'CCOP(=S)(OCC)SCSCC', 'CCCCCCCO',
+  ...          'CCCCCCCC(=O)OC', 'CCOP(=S)(OCC)SCSC(C)(C)C','c3ccc2nc1ccccc1cc2c3', 
+  ...          'Nc2cccc3nc1ccccc1cc23 ', 'C1CCCCCC1','OC1CCCCCC1']
+  >>> Xs= np.zeros(len(data_test))
+  >>> Ys = np.ones(len(data_test))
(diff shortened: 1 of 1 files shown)


davidRFB:
@rbharath  ! I was working in this documentation improvement but I was wondering if it is possible or recommend to add this kind of images in the documentation to explain better some details in the splitter. 
image

image

In addition, the scaffoldSplitter needs the smiles strings in the ids field during the DiskDataset creation : 
```import deepchem as dc
data_test = ['CC(C)=CCCC(C)=CC(=O)', 'CCCC=C', 'CCCCCCCCCCCCCC', 'CC(C)Cl',
           'CCC(C)CO', 'CCCCCCCCCC(C)O', 'CCOP(=S)(OCC)SCSCC', 'CCCCCCCO',
           'CCCCCCCC(=O)OC', 'CCOP(=S)(OCC)SCSC(C)(C)C','c3ccc2nc1ccccc1cc2c3', 
           'Nc2cccc3nc1ccccc1cc23 ', 'C1CCCCCC1','OC1CCCCCC1']
Xs= np.zeros(len(data_test))
Ys = np.ones(len(data_test))
dataset = dc.data.DiskDataset.from_numpy(X=Xs,y=Ys,w=np.zeros(len(data_test)),ids=data_test)
```
this should be specified somewhere else in documentation to clarify this condition ?

rbharath:
@davidRFB Thanks for the contribution! It looks like the flake8/doctests are failing. Can you run flake8 on the file and fix the formatting issues on the docs?

I don't think we can add images but I think adding details about the need for the smiles field would be great

davidRFB:
Hi @rbharath  ! The flake and the doctest now work fine ! I change a little the text. There is something about the splitter. First generate groups of molecules with the same scaffold, and then sort the groups in descending order by group size. This will cause that the first group contains the biggest number of molecules with the same scaffold and the last ones contains just one molecule. In the split, the train,test and validation datasets takes the groups of molecules also in  descending order,  therefore the test and validation will contain molecules with different scaffold while the train will have a lot of molecules with repeated scaffolds. This could be a bias ?

peastman:
Interesting question.  I don't think it's a bias: any scaffold that appears in the test set does not appear in the training set.  A more subtle question is whether this reduces the training efficiency.  Would it do a better job of extrapolating to new scaffolds if there was more diversity of scaffolds in the training set?  Possibly so, although increasing the diversity in the training set would mean decreasing diversity in the test and validation sets.  And since they're much smaller, it makes sense to put the largest scaffold sets into the training set.  Otherwise, if any single scaffold accounts for more than a few percent of the full dataset, you could get a test set that's mostly just one scaffold.

rbharath:
@davidRFB It looks like yapf is still failing but the flake8/doctest look good! Could you rerun yapf and verify you have version `0.22` of `yapf` locally? You can check the version by running
```
yapf --version
```

davidRFB:
Hi @rbharath  Sorry for the confusion ! I had yapf=0.30 . The change was done. However, some CI test failed. I do not know if there is something with my branch.

rbharath (commented):
LGTM!

I believe the errors are due to an issue with pdbfixer/openmm version mismatch that we need to figure out.

PR #2783, opened 2021-12-07
Removing package pins

arunppsg:
DeepChem has a number of packages in dependencies which are pinned to specific versions. In the long run, it is not good to pin packages. In this PR, I am experimenting with removing those pins. Final goal is to bump tensorflow to 2.7.

### DIFF
--- a/deepchem/models/torch_models/cgcnn.py
+++ b/deepchem/models/torch_models/cgcnn.py
@@ -19,9 +19,9 @@ class CGCNNLayer(nn.Module):
   Examples
   --------
   >>> import deepchem as dc
-  >>> import pymatgen as mg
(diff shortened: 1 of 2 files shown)


arunppsg:
When Numpy is updated to 1.21, I am observing some mypy errors. These were also reported previously in  #2382. These might take some time to fix.

peastman:
I can try to take a look if you want.  I previously did some work trying to fix these errors, but put it aside when we decided to pin numpy instead.

arunppsg:
That will be very helpful @peastman . The main issue is the `mypy` errors as part of the core CI.

rbharath (approved):
LGTM

PR #2799, opened 2021-12-27
Minor fixes

arunppsg:
A couple of minor fixes are made in this PR.

- The invocation of mypy in `main.yml` should be `mypy -m deepchem`. Currently it is `mypy -p deepchem`. The difference is that `-p` flag is used to specify a package. Then, mypy will recursively type check all the submodules and subpackages of that packege. This can sometimes cause downstream errors when one of the subpackages has mypy error. To fix it, we can use `--module` (`-m`) option. This will only type check that specific module. Ref: https://mypy.readthedocs.io/en/stable/running_mypy.html#specifying-code-to-be-checked
- The other fixes are for flaky tests.

### DIFF
--- a/deepchem/models/tests/test_atomic_conv.py
+++ b/deepchem/models/tests/test_atomic_conv.py
@@ -3,9 +3,8 @@ Tests for Atomic Convolutions.
 """
 
 import os
-
 import pytest
-
+from flaky import flaky
 import numpy as np
 from deepchem.data import NumpyDataset
 from deepchem.feat import AtomicConvFeaturizer
@@ -35,6 +34,7 @@ def test_atomic_conv_initialize():
   assert len(acm.atom_types) == 15
 
 
+@flaky
 @pytest.mark.slow
 @pytest.mark.tensorflow
 def test_atomic_conv():
--- a/deepchem/models/tests/test_cnn.py
+++ b/deepchem/models/tests/test_cnn.py
@@ -2,6 +2,7 @@ import deepchem as dc
 import numpy as np
 import pytest
 import unittest
+from flaky import flaky
 try:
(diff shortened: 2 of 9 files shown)


peastman:
Isn't `-p` what we want?  It *should* check all submodules inside `deepchem`.

arunppsg:
I came across a [CI error](https://github.com/deepchem/deepchem/runs/4636293709?check_suite_focus=true#step:15:12) yesterday. `ruamel.yaml` was updated with an incorrect type recently which was producing this error ([ref](https://sourceforge.net/p/ruamel-yaml/tickets/412/)). Ideally, mypy check should run only on `deepchem` and its submodules.

> Isn't `-p` what we want? It _should_ check all submodules inside `deepchem`.

mypy docs states that `-p` flag recursively type checks a package for submodules. I read mypy docs once again and I agree that `-m` option does not check recursively.

What I am not able to explain here is why we are getting a mypy error on package `ruamel` which we are not using at all. Also, I noticed that just using `mypy deepchem` fixes the CI error ([ref](https://github.com/deepchem/deepchem/runs/4640000025?check_suite_focus=true#step:15:12)).

peastman:
I've never seen that error.  Are you able to reproduce it locally?  From the log, it looks like that package gets downloaded as part of installing miniconda.

arunppsg:
I can't reproduce it locally as the installation of deepchem using the installation scripts fails in mac M1. I use packages installed individually. This is another thing which is to be fixed - we have build scripts for mac OS failing too.

arunppsg:
The mypy error vanished. Hence, dropping that commit.

arunppsg:
CI is passing and will merge this in!

PR #2807, opened 2022-01-09
Minor fixes

arunppsg:
Fixed are introduced for:

- macOS build CI - torch installation in macOS to use `pytorch` conda channel here after to avoid segmentation fault when importing
- Removing two redundant installation files: `env_torch.yml` and `env_jax.yml`. Already existing `env_common.yml` can be used instead of the two


arunppsg:
Summary:

- This PR mainly fixes macOS build CI
- It also find issues with windows OS for common and core CI

The three failures are:

- Common CI windows 3.7 - failures in pytest for vina on windows since vina is not supported with windows. We need to add probably `windows` tag in pytest and exclude those
- Core CI windows 3.7 are due to failures due to jax as jax is not supported in windows.

arunppsg:
@rbharath can you review this one when you are available and merge it in?

rbharath (approved):
LGTM!

PR #2850, opened 2022-02-04
Gaussian processes tutorial

TheRealSalmon:
# Pull Request Template

## Description

Added tutorial for using Sklearn's Gaussian Processes. Also updated the readme table to contents to include this new tutorial. 

## Type of change

## Checklist


peastman:
This is a really nice tutorial!

Could you add a brief description of Gaussian processes at the beginning?  Assume many readers aren't familiar with them.  Linking to papers for full details is good, but if you could add just one paragraph very briefly describing what they are and what they're useful for, that would be very helpful.

In the discussion of uncertainty, it would be useful to refer back to the concepts introduced in the "Uncertainty in Deep Learning" tutorial.  I believe the values you compute include only aleatoric uncertainty, no epistemic uncertainty.  Is that correct?

> Unfortunately we find that there turns out to be no correlation between the observed error and the predicted standard deviation. This is to be expected for GP.

This needs a *lot* more discussion!  Why is it expected?  The title of the tutorial is, "Uncertainty Estimation using Gaussian Processes," and you spend a lot of time estimating uncertainties.  And then at the very end you discover your uncertainty estimates are meaningless, and tell us that's to be expected!  Then what was the point of the tutorial?

TheRealSalmon:
Hi Peter,

Thanks for taking the time to look at my work. Upon further reading, I think my original conclusions were wrong. I will update my work and present my new conclusion here.

The GP model I created seems to only capture homoskedastic aleatoric uncertainty as the uncertainty is uniform and is ~0.7 log units which seems reasonable for typical error in a biological dataset.

However, recent work using MPNN-GP (see https://arxiv.org/abs/2005.10036) seems to suggest that GPs are in fact capable of capturing epistemic uncertainty.

I think the reason why MPNN-GP is capable of capturing epistemic uncertainty is the feature vector produced by the trained MPNN model is information dense and the distance between vectors is "meaningful" when it comes to uncertainty.

However, the feature vector generated by ECFP is less information dense, and as it is one-hot encoded, the distances between ECFP vectors may be less meaningful when it comes to uncertainty.

I am honestly still a beginner in this field so I would definitely appreciate any help you have with a more reasonable conclusion. You're right though, I shouldn't have given such a cop-out conclusion!

Best,
Sam

PR #2855, opened 2022-02-10
Removing some obsolete code

arunppsg:
In this PR, I have removed some old data files which are not used now and also updated sphinx configuration.


arunppsg:
The data files I deleted were from `deepchem/models/tests` directory. These files are not used by the test files in models/tests but it seems that they are used somewhere else (in `deepchem/trans/tests`). I will restore the files to `trans/tests/assets` directory as their used there.

arunppsg:
Broadly, in this PR I cleaned `models/tests` directory by moving data files into `models/tests/assets` and removing data files which were not used any more.

PR #2864, opened 2022-02-18
Fresh gp tutorial

TheRealSalmon:
# Pull Request Template

## Description

Added a tutorial showing how to use Gaussian Processes in DeepChem.

## Type of change

## Checklist


TheRealSalmon:
@peastman I have added some more content in response to your comments. Do you think the discussion is sufficient? Thank you!

peastman:
> A further point of discussion has to do with aleatoric uncertainty (uncertainty in the measurement) and epistemic uncertainty (uncertainty in the model's prediction).

That's still the wrong description of epistemic uncertainty.  How about, "uncertainty about what is the correct model".

Your added calculations let us draw some more conclusions.  If it were correctly predicting the overall magnitude of the uncertainty, 68% of samples would be within one standard deviation.  Instead you get 73% on the test set (only slightly too high) and 94% on the training set (way too high).

Whenever you see a model doing better on the test set than the training set, you know something is very wrong!  Basically, it's done a poor job of learning the uncertainty in the training data.  It predicts uncertainties about 2x too large.  When you move to the test set, it still gives uncertainties that are too large, but because its predictions happen to be less accurate, they're not as badly too large.  That's just a coincidence though.

If you believe the problem comes from the use of ECFP as the featurization, what about trying a different example using a featurization you think would be more appropriate?

One minor detail: you show plots of measured vs. predicted values for the training set and test set.  They're supposed to show that the correlation is less good for the test set, but because the plots have different axis ranges, it's hard to really compare them.  Fixing both of them to use the same ranges would make it easier to see how they really compare.

TheRealSalmon:
Hi Peter, 

I will respond to each of your points paragraph by paragraph:

I will update the definition of epistemic uncertainty, thanks for that.

I see that the error on the test set is close but that it strongly overpredicts error on the training set. That is a good point of discussion and I will add it to the tutorial.

I would personally push back on you saying that something is very wrong. Overfitting to the training set is a problem that all models suffer from. Rather, I'm more concerned that the model is not learning uncertainty well. 

On that point, I tried to use Mol2Vec as a featurizer thinking that using a more information-rich feature space would help with the uncertainty. But after ~1 hour of trying to get it to work, I couldn't get the featurizer to work. I believe it has something to do with the version of gensim I'm using. The mol2vec github does not specify what version of gensim it uses. Furthermore, the older version to gensim seem to be incompatible with the version of scipy I have installed with deepchem. 

I will also fix the axes, that's a good point.

Okay now that I've gone through all your points, I do recognize the tutorial does kind of suck. I was really hoping Gaussian Processes would just work out of the box with a standard fingerprint. As I point out in the tutorial, multiple academic papers have been successful implementing a MPNN-GP model where the MPNN (chemprop) is trained on some of the training set and the GP is trained on the MPNN latent space as a fingerprint on the rest of the training set. I think that this is pretty cutting edge and honestly a little outside the scope of the tutorial. I also don't really want to be doing other people's "homework" and would rather leave it to them to try to extend beyond what the tutorial shows. 

I agree that the tutorial leaves much to be desired but honestly I'd rather leave it as it is and work on other things within deepchem. If you feel that the quality is not good enough to be published as a tutorial, I understand. I'd probably want to revisit it once chemprop is wrapped or implemented in deepchem or later when I have some more energy to deal with the headache surrounding mol2vec, scipy, and gensim versions. 

No matter what you decide, I'm thankful for your discussions and help, I learned the most making this tutorial. 

Thanks,
Sam

peastman:
> Overfitting to the training set is a problem that all models suffer from. Rather, I'm more concerned that the model is not learning uncertainty well.

My point was just that the better performance on the test set is only a coincidence.  The model only knows about the data it was trained on, not the data it wasn't trained on.  If you find better performance on unseen data than on training data, you can be pretty certain it's from a cancellation of errors, not any good reason.  It's bad at predicting uncertainties, but on the test set the overall magnitude of the actual errors just happens to be closer to its predicted uncertainties.

I wish we could find one example where it clearly works.  If we can't, maybe we should just rework it to put less emphasis on uncertainty prediction.  GaussianProcessRegressor is still a fine model, and a tutorial showing how to use it is useful.  Then near the end you can say, "By the way, this model can also be used to predict uncertainties.  Here's how you do it.  It doesn't seem to be doing a very good job, but other authors have reported better results.  Here's the reference."  That just shouldn't be the main focus of the tutorial.

By the way, in the paper they did try using fingerprints (see the results for FP GP).  They found the results were not quite as good as MPNN GP, but they were close.  That isn't the reason it isn't working here.

TheRealSalmon:
@peastman I read the paper again and saw that they used the Linear kernel instead of the RBF kernel like I did. I got really excited to try the change! But when I tried scikit-learn's DotProduct kernel with sigma_0 = 0 (equivalent to a Linear kernel) I didn't see any improvement in the uncertainty calibration. That's too bad!

I think as we've discussed, we can pivot the current tutorial to just be an example of how to use scikit-learn's Gaussian Processes. We can revisit the uncertainty estimation part of the tutorial later. Let me know what you think.

peastman:
Oh well, it was a good try!  I agree with your suggestion.

TheRealSalmon:
@peastman I have slimmed down the tutorial by cutting it short. What do you think, do you think the scope is more appropriate given the results?

peastman:
This looks great.  The focus now fits well with the results.

> For the train set, >90% of the samples are within a standard deviation. In comparison, only ~70% of the samples are within a standard deviation. 

I think there's a missing "for the test set" in the second sentence?  Or do you mean, "for a normal distribution"?

TheRealSalmon:
Okay I made the change.

peastman:
I think this is ready as far as I'm concerned.

rbharath (approved):
LGTM.

Congats on your first merged DeepChem PR @TheRealSalmon!

PR #2873, opened 2022-02-25
removing package pin

arunppsg:
Changes made:
- removed pin for yapf

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
       - id: check-yaml
 
 -   repo: https://github.com/pre-commit/mirrors-yapf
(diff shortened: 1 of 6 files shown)


peastman:
If we remove the pin, no one will know which version they need to use, and every time a new version comes out everyone's PRs will immediately start giving errors.  I suggest we continue to pin it, just to a newer version.

arunppsg:
Makes sense, I will pin it to the latest version of yapf (0.32.0).

arunppsg:
Yapf version bump does not seem to cause additional failures in CI.

arunppsg:
Merging this in as the 2 observed failures in CI are due to the usual ones (jax in windows).

PR #2914, opened 2022-05-29
Adding DeepQMC tutorial

shaipranesh2:
# Pull Request Template

## Description

Fix #(issue)

As part of issue #2907 , differential quantum chemistry tutorial- DeepQMC tutorial has been added,

## Type of change

## Checklist


tonydavis629:
My suggestions to clean this up and improve this tutorial:
- Remove the outputs of the keyboard interrupts, no need to start the training for C2H6 because it will not be able to train in colab anyway. But I think it's valuable to show how it might be used given enough resources.
-import Molecule from deepqmc in the hydrogen molecule section so it displays the molecules as expected.
-include a little more detail in the background section to help introduce the idea of quantum chemistry to someone who may not be familiar. Maybe a simple graphic of electron orbitals would remind people of their chemistry classes.
-"This tutorial and the rest in this sequence..." part is duplicated, remove the extra 
- Peter's suggestions on solving a slightly more interesting problem and graphically representing the results, such as H2+ as a function of the proton distance.  

Overall good progress! I think this tutorial does a good job displaying how to use paulinet really well. Once you have ferminet implemented, you can add it to this tutorial and make it a more general "quantum chemistry in deepchem" tutorial. 

There are some more CI failures than expected but they don't seem to be related to this PR.

peastman:
When writing a tutorial, you want to start by asking a few questions.  Who is the audience for the tutorial?  What do you assume they already know?  What do you want them to learn from reading it?  Anything you want them to learn needs to be explained in enough detail for them to understand it.  Anything else is a distraction and better left out.  In this case, I think we want to assume they have studied quantum mechanics, but don't assume they know anything about quantum chemistry methods?  The goal of the introduction is to teach them just enough about it to understand what you're doing in the rest of the tutorial.

I think these are the main points you want to make:

- The state of a molecule is described by its wavefunction.
- A lot of important properties are determined by the ground state, that is, the state with lowest energy.  Determining the ground state is an important problem for a lot of applications.
- The atomic nuclei are much heavier and slower moving than the electrons, so usually we just treat them as fixed.  Then we solve for the wavefunction of the electrons in the presence of the fixed nuclei.
- Lots of methods have been developed for doing that.  They make different tradeoffs between speed and accuracy.  They range from really fast but inaccurate to really accurate but incredibly slow.
- Neural networks can approximate any function, so why not a wavefunction?  It's just a function that computes a complex number given the positions of the electrons.  We can optimize the model parameters to minimize the energy, which will give an approximation for the ground state.
- PauliNet is a neural network architecture that's specifically designed to be good at representing wavefunctions.

It's best to avoid technical jargon (Slater determinants, Hartree-Fock, etc.) unless you're specifically going to make use of them later in the tutorial (in which case they need to be explained).

peastman:
> Peter's suggestions on solving a slightly more interesting problem and graphically representing the results, such as H2+ as a function of the proton distance.

I was actually thinking of H2, unless leaving off an electron is necessary to make the computation time practical.  H2+ is less interesting in this case.  I mean, the whole point of PauliNet is that it's antisymmetric under exchange of electrons.  If you only have one electron, there's nothing to exchange!

peastman:
The introduction looks a lot better!  Here are a few minor changes.

> This probability is given by the square of determinant of a property of molecule system called the wavefunction

"squared magnitude", not "square of determinant".  The value of the wavefunction is just a complex number, not a matrix.  It doesn't have a determinant.

> The nucleus of the atoms in a molecule are considered to be stationary and then we solve to get the wavefunction.

To make it a little clearer: "For many purposes, the nuclei of the atoms in a molecule can be considered to be stationary, and then we solve to get the wavefunction of the electrons."

> but for heavier atoms and molecules, electron-electron repulsion and nucleus-nucleus repulsion comes into act and makes it hard to compute the wavefunctions due to these interactions.

That should be, "electron-electron repulsion and electron-nucleus attraction".  We're treating the nuclei as fixed and just solving for the wavefunction of the electrons, so repulsion between nuclei doesn't come into it.

Did you figure out a way to make the calculation faster?  How long does each energy take?

The code for evaluating all the energies could be made simpler and clearer by putting it in a loop.  I'm thinking of something like this.

```python
distances = [0.4, 0.5, 0.6, 0.7, 0.9, 1.1, 1.3, 1.5]
energies = []
for distance in distances:
    mol = Molecule(
        coords=[[0, 0, 0], [distance, 0, 0]],
        charges=[1, 1],
        charge=0,
        spin=0,
    )
    net = PauliNet.from_hf(mol).cuda()
    train(net, n_steps=1000, batch_size=1000, epoch_size=50)
    molecule = evaluate(net)
    energies.append(uncertainties.nominal_value(molecule['energy']))

**Comment by peastman** [comment]
I was curious to see how accurate the results are, so I computed the same energies with Psi4 using a high level of theory and a large basis set.

```python
import psi4

distances = [0.4, 0.5, 0.6, 0.7, 0.9, 1.1, 1.3, 1.5]
energies = []
for distance in distances:
    mol = psi4.geometry(f"""
    0 1
    H 0 0 0
    H 0 0 {distance}
    """)
    energies.append(psi4.energy('ccsd(t)/cc-pVQZ', molecule=mol))
print(energies)
```

It only took a fraction of a second for each calculation.  Here are the results:

```
[-0.9808444102252339, -1.1033812202002902, -1.1553104904932252, -1.1725277483650176, -1.1617826395442494, -1.1304539990023956, -1.0968985131662812, -1.0675101778667093]
```

And here's what you get for `hf/6-31G`, a much less accurate level of theory with a much smaller basis set (which also took only a fraction of a second for each calculation):

```
[-0.933007362411975, -1.0580271105758703, -1.1100324287663494, -1.1261243205408276, -1.1116878047525698, -1.0756881891665624, -1.0358298517451476, -0.9975015291244915]
```

These are variational methods, so lower values are always better.  Comparing to the DeepQMC results, I'd say it's... not too good. :)  I guess this is still a research project!

shaipranesh2:
> These are variational methods, so lower values are always better. Comparing to the DeepQMC results, I'd say it's... not too good. :) I guess this is still a research project!

Yea, only when I ran for a long time(like some 40 mins), I got close to these results. I tried locally running, but again got same results. I guess it has still a long way to go and paulinet is supposed to be faster than ferminet!!!

shaipranesh2:
> It only took a fraction of a second for each calculation. Here are the results:
> 
> ```
> [-0.9808444102252339, -1.1033812202002902, -1.1553104904932252, -1.1725277483650176, -1.1617826395442494, -1.1304539990023956, -1.0968985131662812, -1.0675101778667093]
> ```
> 
> And here's what you get for `hf/6-31G`, a much less accurate level of theory with a much smaller basis set (which also took only a fraction of a second for each calculation):
> 
> ```
> [-0.933007362411975, -1.0580271105758703, -1.1100324287663494, -1.1261243205408276, -1.1116878047525698, -1.0756881891665624, -1.0358298517451476, -0.9975015291244915]
> ```
> 
> These are variational methods, so lower values are always better. Comparing to the DeepQMC results, I'd say it's... not too good. :) I guess this is still a research project!

😅 actually i took a look into DeepQMC again, the coordinates should be mentioned in terms of bohrs to get the solution, which was little vague in their docs and found only after going through their code. Now i got the solution similar as above like within a minute or two.

peastman:
Units are the source of so many problems!

peastman:
That graph looks much better now!  The text still says, "It shoud contain the list of coordinates of each nucleon in angstroms(coord)," so that needs to be fixed.

This line is a bit confusing:

> `distances = [0.4*bohr, 0.5*bohr, 0.6*bohr, 0.7*bohr, 0.9*bohr, 1.1*bohr, 1.3*bohr, 1.5*bohr]`

I would read that as, "We're computing distances of 0.4 bohr, 0.5 bohr, etc."  But actually those distances are angstroms.  Maybe rename the variable from `bohr` to `angstroms_to_bohr`?  You also could make it simpler by doing the multiplication only once:

```python
distances = [0.4, 0.5, 0.6, 0.7, 0.9, 1.1, 1.3, 1.5] # distances in angstroms
```

And then

```python
coords=[[0, 0, 0], [distance*angstroms_to_bohr, 0, 0]],
```

tonydavis629:
This tutorial is ready to merge

PR #2931, opened 2022-06-04
Fixing CI errors

arunppsg:
Changes in this PR:
- The torch CI has been failing for some time due to an import error in `tensorboard`. tensorboard team made a fix which was available in the nightly version, hence updated requirements to use tensorboard nightly version. This should be reverted back to tb stable when tb 2.9.1/2.10.0 is released.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7]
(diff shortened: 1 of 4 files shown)


arunppsg:
Other changes in this PR:
- In `test_compute_pairwise_distance`, the distance should be less than `sqrt(3)` since it is a 3D space.
- Removed version pin on `sphinx` and `sphinx_rtd_theme`
- Updated python version to 3.8 for docs build

PR #2935, opened 2022-06-07
Adding Metropolis Hasting sampler

shaipranesh2:
# Pull Request Template

This PR aims to add MCMC sampling class to deepchem for later use in Ferminet

Fix #(issue)

## Type of change

## Checklist

### DIFF
--- a/deepchem/utils/ElectronSampler.py
+++ b/deepchem/utils/ElectronSampler.py
@@ -2,7 +2,7 @@ import numpy as np
 from typing import Optional, Callable
 
 
-class MCMC:
+class ElectronSampler:
   """
   This class enables to initialize electron's position using gauss distribution around a nucleus and update using Markov Chain Monte-Carlo(MCMC) moves.
 
@@ -22,39 +22,40 @@ class MCMC:
 
   Example
   -------
-  >>> from deepchem.utils.ElectronSampler import MCMC
+  >>> from deepchem.utils.ElectronSampler import ElectronSampler
   >>> def test_f(x):
   ...	#dummy function which can be passed as the parameter f. f gives the log probability
   ...	return 2*np.log(np.random.uniform(low=0,high=1.0,size=np.shape(x)[0]))
-  >>> distribution=MCMC(central_value=[[1,1,3],[3,2,3]]],f=test_f,seed=0,batch_no=2,steps=1000,)
-  >>> distrubution.gauss_initialize_position([[1],[2]])
-  >>> distribution.x
-  [[[[1.00626135 0.98291809 2.9489402 ]]
+  >>> distribution=ElectronSampler(central_value=np.array([[1,1,3],[3,2,3]]),f=test_f,seed=0,batch_no=2,steps=1000,)
+  >>> distribution.gauss_initialize_position(np.array([[1],[2]]))
+  >> print(distribution.x)
+  [[[[1.03528105 1.00800314 3.01957476]]
 
-    [[3.04539509 1.97091269 3.00091517]]
+    [[3.01900177 1.99697286 2.99793562]]
 
-    [[2.99625632 2.03065558 3.02938718]]]
+    [[3.00821197 2.00288087 3.02908547]]]
 
 
-   [[[1.01307237 1.01728872 2.9851567 ]]
+   [[[1.04481786 1.03735116 2.98045444]]
 
-    [[3.00309895 2.00756325 2.98224429]]
+    [[3.01522075 2.0024335  3.00887726]]
 
-    [[2.96038407 1.99304176 3.00312698]]]]
+    [[3.00667349 2.02988158 2.99589683]]]]
   >>> distribution.move()
-  >>> ditribution.x
-  [[[[-0.08888387  1.24021524  2.77555481]]
+  0.5115
+  >> print(distribution.x)
+  [[[[-0.32441754  1.23330263  2.67927645]]
 
-    [[ 3.34737279  2.27259311  3.52041236]]
+    [[ 3.42250997  2.23617126  3.55806632]]
 
-    [[ 3.26025284  1.71305735  3.1379624 ]]]
+    [[ 3.37491385  1.54374006  3.13575241]]]
 
 
-   [[[ 0.43672637  0.96475675  3.7220207 ]]
+   [[[ 0.49067726  1.03987841  3.70277884]]
 
-    [[ 3.30810994  1.58044327  2.56373753]]
+    [[ 3.5631939   1.68703947  2.5685874 ]]
 
-    [[ 2.63144407  1.89442955  3.62751124]]]]
+    [[ 2.84560249  1.73998364  3.41274181]]]]
   """
 
   def __init__(self,
@@ -171,7 +172,7 @@ class MCMC:
            stddev: float = 0.02,
            asymmetric_func: Optional[Callable[[np.ndarray], np.ndarray]] = None,
            index: Optional[int] = None) -> float:
-    """Performs Metropolis-Hasting move for self.x(electrons). The type of moves to be followed -(simultaneous or single-elctron, symmetric or asymmetric) have been specified when calling the class.
+    """Performs Metropolis-Hasting move for self.x(electrons). The type of moves to be followed -(simultaneous or single-electron, symmetric or asymmetric) have been specified when calling the class.
     The self.x array is replaced with a new array at the end of each step containing the new electron's positions.
 
     Parameters:
--- a/deepchem/utils/ElectronSampler.py
+++ b/deepchem/utils/ElectronSampler.py
@@ -1,0 +1,265 @@
+import numpy as np
+from typing import Optional, Callable
+
+
+class MCMC:
+  """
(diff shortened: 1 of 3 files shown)


peastman:
Looks like a good start!  Here are a few general comments.  The docs page on [coding conventions](https://deepchem.readthedocs.io/en/latest/development_guide/coding.html) gives lots more information about some of them.

Docstrings should use Numpy style.  For example,

```
steps: Integer, the number of steps performed when called
```

should instead be

```
steps: int
    the number of steps performed when called
```

Type annotations are needed on arguments and return values.

Could you add a top level docstring for the class?  It should describe what the class is for, cite the algorithms it uses, and provide a usage example.

You'll need to run yapf on the code to make the tests pass.

Can you add unit tests for the new code?

The class name MCMC is too general.  This class does something much more specific: MCMC on electron positions by applying displacements drawn from a Gaussian distribution.  Perhaps something like ElectronPositionSampler would be a better fit?

tonydavis629:
Per the 6/17 discussion:
I would say don't worry about the single electron move if it is not a trivial task. It is more important to get the overall project working first and then you can always go back and add in extra options in the sampler.

Usage examples will help you write code as you want it. I recommend in the future, before you start writing the code, write some examples of how you want to use the code and include that in the docstring at least for the class.  I think if you add some example usage in the class docstring then the documentation would be good to go. They will also help you write the unit tests by just following your examples.

tonydavis629:
Notes from our 6/19 meeting:

Since this sampler doesn't work as a general MC sampler you will need to change the name to electron sampler or something along those lines. In the future we can add an abstract MCMC class and this will be derived from that, but that's for another time. 

In the testing, it looks like you still call the simultaneous move function when your comments indicate testing the single move. 

In order to get the doc test to succeed in the CI, we need to make sure the doctest is either not run on lines that have some kind of stochastic process like MC, or we need to set the random number generator seed to a certain value so that when such a process is run by the CI that it ends up with the same answer as the doc string example. I would probably recommend setting the seed so you can still run the tests. But otherwise, to add an example to a docstring without it having run in the doc test, you simply use >> instead of >>> to indicate code that should be included in documentation but not in doc testing. 

As far as your question on implementing the loss, l see in the paper it uses the difference between the network output and Hartree Fork. So you will need to implement that in pytorch in order to input to the optimizer in order to train the network. Not sure if that answers your question.

Looking ahead to the next steps, I would recommend contacting the author of https://github.com/alecwangcq/KFAC-Pytorch to get his permission to use his KFAC pytorch implementation in our MIT codebase. And of course cite the author in the docstrings when including it in deepchem.

peastman:
> Since this sampler doesn't work as a general MC sampler you will need to change the name to electron sampler or something along those lines.

Along the same lines, make the docstring for the class more detailed to explain exactly what problem it's used for.

> I would probably recommend setting the seed so you can still run the tests.

See [CoulombMatrix](https://github.com/deepchem/deepchem/blob/master/deepchem/feat/molecule_featurizers/coulomb_matrices.py) for an example of a class that takes a random seed as an optional argument.

deepchem/utils/monte_carlo.py
```diff
@@ -0,0 +1,251 @@
+import numpy as np
+
+
+class MCMC:
+  """
+	Performs Markov chain Monte Carlo(MCMC) steps on data samples which are provided. It can also initialize data around a mean via gaussian distribution.
+	Has 2 types of moves - simultaneous and specific moves.
```
peastman, line 7: Can you explain what the difference is between them?

peastman (commented):
I had to study the code to understand the big picture of how this class works.  These seem to be the basic principles:

- Internally it stores an array `x` containing the current state of whatever system you're modelling with Monte Carlo.
- The initial value of it can optionally be passed to the constructor.
- Each time you call one of the "move" or "initialize" methods, it creates a new `x` array, replacing the old one.

It would be nice to describe this more clearly.  Be sure to emphasize that it always creates a new array instead of modifying the old one, since that can easily lead to confusion.  For example, if you write

```python
x = mcmc.x
mcmc.simultaneous_move(...)
```

then `x` will not be changed, and will no longer equal `mcmc.x`.

It might be worth thinking a bit about the API for moves.  There are a lot of options that affect how moves get performed.  They get specified in several different ways:

- The number of steps to perform on each move is specified once as an argument to the constructor.
- The choice of whether to perform single or simultaneous moves is determined by which method you call.
- The choice of whether to perform symmetric or asymmetric moves and the stddev of the moves are determined by arguments passed to the move methods.

Is there a good reason to specify different options in different ways?  For example, would it make more sense to turn all of these into constructor arguments and have a single `move()` method that takes few or no arguments?  Or alternatively to turn all of them into arguments of a single `move()` method?

deepchem/utils/monte_carlo.py
```diff
@@ -0,0 +1,251 @@
...
+	"""
+
+  def __init__(self, central_value, batch_no=10, x=[], steps=10):
+    """
+		Parameters:
+		-----------
+		central_value: List, 
+		List of integer,acts as the mean to initial distribution of data. Should be in the form [[1,2,..],[3,4,..],..]
+		batch_no: Integer, default 10
+		Number of batches for the data to be initialized
```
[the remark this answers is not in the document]
shaipranesh2: The batch number here I meant was the number of batches of the electrons to take when initialized.

deepchem/utils/monte_carlo.py
```diff
@@ -0,0 +1,251 @@
...
+		x: List, default []
+		List of integer. The data distribution for which the MCMC should be run. Should be in the shape (batch_no,number_of_samples,1,dimension_of_data)
```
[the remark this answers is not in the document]
shaipranesh2: I thought to use it for a general MCMC case, but should change this with respect to the context of electrons and change it electron coordinates. And yes, it should be an array here

shaipranesh2:
> It would be nice to describe this more clearly. Be sure to emphasize that it always creates a new array instead of modifying the old one, since that can easily lead to confusion. For example, if you write
> 
> ```python
> x = mcmc.x
> mcmc.simultaneous_move(...)
> ```
> 
> then `x` will not be changed, and will no longer equal `mcmc.x`.

right now, this code updates the current state array x. Is it fine if it continued to be like this? (as I thought that the history of moves is not really necessary otherwise the user can log it on his own using small moves)

peastman:
> right now, this code updates the current state array x.

I don't think so?  It looks like all the functions create a new `x` array rather than modifying the existing one in place.  For example,

```python
self.x = np.where(cond[:, None, None, None], x2, self.x)
```

If you wanted to update the array in place, the code would be something like

```python
self.x[cond[:, None, None, None]] = x2[cond[:, None, None, None]]
```

The important thing is to clearly document which it does.

deepchem/utils/ElectronSampler.py
```diff
@@ -0,0 +1,266 @@
...
+
+  Example
+  -------
+  >>> from deepchem.utils.ElectronSampler import MCMC
+  >>> def test_f(x):
+  ...	#dummy function which can be passed as the parameter f. f gives the log probability
+  ...	return 2*np.log(np.random.uniform(low=0,high=1.0,size=np.shape(x)[0]))
+  >>> distribution=MCMC(central_value=np.array([[1,1,3],[3,2,3]]),f=test_f,seed=0,batch_no=2,steps=1000,)
+  >>> distribution.gauss_initialize_position(np.array([[1],[2]]))
+  >>> print(distribution.x)
```
tonydavis629, line 31: Use >> instead of >>> to not trigger the doc test

docs/source/api_reference/utils.rst
```diff
@@ -237,3 +237,13 @@ used for testing model architectures or other purposes.
...
   :members:
+
+ElectronSampler
+-------------------
+
+The utilities here are used to sample electrons in a given molecule
+and update it using monte carlo methods, which can be used for methods
+like Variational Monte Carlo, etc.
+
+.. autoclass:: deepchem.utils.ElectronSampler.MCMC
```
tonydavis629, line 248: Make sure to build the docs on your local machine to make sure that the documentation loads correctly. docs/readme.md has some instruction on how to do that.

deepchem/utils/ElectronSampler.py
```diff
@@ -0,0 +1,266 @@
...
+          self.x,
+          np.random.normal(mean, stddev,
+                           (self.batch_no, specific_sample, 1, ndim)),
+          axis=1)
+
+  def move(self,
+           stddev: float = 0.02,
+           asymmetric_func: Optional[Callable[[np.ndarray], np.ndarray]] = None,
+           index: Optional[int] = None) -> float:
+    """Performs Metropolis-Hasting move for self.x(electrons). The type of moves to be followed -(simultaneous or single-elctron, symmetric or asymmetric) have been specified when calling the class.
```
tonydavis629, line 175: > single-elctron 

spelling

deepchem/utils/ElectronSampler.py
```diff
@@ -0,0 +1,266 @@
+import numpy as np
+from typing import Optional, Callable
+
+
+class MCMC:
```
peastman, line 5: Remember to rename the class as well as the file.

peastman (approved):

PR #2972, opened 2022-07-01
Adding KFAC

shaipranesh2:
# Pull Request Template

## Description

Fix #(issue)

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/optimizers.py
+++ b/deepchem/models/optimizers.py
@@ -684,15 +684,35 @@ class KFAC(Optimizer):
 
   def __init__(self, **kwargs):
     """
-    Parameters
-    ----------
-    **kwargs: 
+    Parameters:
+    -----------
+    model: torch.nn.Module
+    The model to be optimized.
+    lr: float (default: 0.001)
+    Learning rate for the optimizer.
+    momentum: float (default: 0.9)
+    Momentum for the optimizer.
+    stat_decay: float (default: 0.95)
+    Decay rate for the update of covariance matrix with mean.
+    damping: float (default: 0.001)
+    damping factor for the update of covariance matrix.
+    kl_clip: float (default: 0.001)
+    Clipping value for the update of covariance matrix.
+    weight_decay: float (default: 0)
+    weight decay for the optimizer.
+    Tcov: int (default: 10)
+    The number of steps to update the covariance matrix.
+    Tinv: int (default: 100)
+    The number of steps to calculate the inverse of covariance matrix.
+    batch_averaged: bool (default: True)
+    States whether to use batch averaged covariance matrix.
+    mean: bool (default: False)
+    States whether to use mean centered covariance matrix.
     """
     self.kwargs = kwargs
 
-  def _create_pytorch_optimizer(self, model):
+  def _create_pytorch_optimizer(self):
     from deepchem.models.torch_models.kfac_optimizer import KFAC_Optimizer
-    self.kwargs['model'] = model
     if isinstance(self.learning_rate, LearningRateSchedule):
       self.kwargs['lr'] = self.learning_rate.initial_rate
     else:
--- a/deepchem/models/optimizers.py
+++ b/deepchem/models/optimizers.py
@@ -690,11 +690,11 @@ class KFAC(Optimizer):
     """
     self.kwargs = kwargs
 
-  def _create_pytorch_optimizer(self, params):
+  def _create_pytorch_optimizer(self, model):
     from deepchem.models.torch_models.kfac_optimizer import KFACOptimizer
-    self.kwargs['params'] = params
+    self.kwargs['model'] = model
     if isinstance(self.learning_rate, LearningRateSchedule):
-      lr = self.learning_rate.initial_rate
+      self.kwargs['lr'] = self.learning_rate.initial_rate
     else:
-      lr = self.learning_rate
+      self.kwargs['lr'] = self.learning_rate
     return KFACOptimizer([self.kwargs])
--- a/deepchem/models/optimizers.py
+++ b/deepchem/models/optimizers.py
@@ -688,13 +688,13 @@ class KFAC(Optimizer):
     ----------
     **kwargs: 
     """
-    self.kwargs=**kwargs
+    self.kwargs = kwargs
 
   def _create_pytorch_optimizer(self, params):
     from deepchem.torch_models.kfac_optimizer import KFACOptimizer
-    self.kwargs['params']=params
+    self.kwargs['params'] = params
     if isinstance(self.learning_rate, LearningRateSchedule):
       lr = self.learning_rate.initial_rate
     else:
       lr = self.learning_rate
-    return KFACOptimizer([**self.kwargs])
+    return KFACOptimizer([self.kwargs])
--- a/deepchem/models/optimizers.py
+++ b/deepchem/models/optimizers.py
@@ -687,27 +687,27 @@ class KFAC(Optimizer):
     Parameters:
     -----------
     model: torch.nn.Module
-    The model to be optimized.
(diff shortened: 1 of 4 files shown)


peastman:
Since the KFAC code was copied from https://github.com/alecwangcq/KFAC-Pytorch, we should add a license header at the top of the file.  Credit the original author, give a link to the original repository, and state the exact license terms he gave you permission to use it under.

deepchem/models/torch_models/torch_model.py
```diff
@@ -1,5 +1,9 @@
 import numpy as np
-import torch
```
arunppsg, line 2: We should preserve this `import torch` statement here.

deepchem/models/tests/test_optimizers.py
```diff
@@ -1,5 +1,7 @@
+import deepchem as dc
 import deepchem.models.optimizers as optimizers
 import unittest
+from deepchem.models.torch_models import kfac_optimizer
```
arunppsg, line 4: This import statement depends on torch module. It is better to keep it's scope local to the test method. In this case, it will `test_KFAC` method

deepchem/models/torch_models/torch_model.py
```diff
@@ -945,10 +953,10 @@ def default_generator(
...
     """
     for epoch in range(epochs):
-      for (X_b, y_b, w_b, ids_b) in dataset.iterbatches(
-          batch_size=self.batch_size,
-          deterministic=deterministic,
-          pad_batches=pad_batches):
+      for (X_b, y_b, w_b,
+           ids_b) in dataset.iterbatches(batch_size=self.batch_size,
+                                         deterministic=deterministic,
+                                         pad_batches=pad_batches):
```
tonydavis629, line 959: Was this done by yapf? Doesn't seem right. I'm not sure if all these formatting changes are necessary.

deepchem/models/tests/test_optimizers.py
```diff
@@ -1,3 +1,5 @@
+from haiku import Linear
```
peastman, line 1: Did you intend to add this?  I don't think you use it anywhere.  Since it requires JAX, it should only be imported inside functions that are marked for JAX only.

shaipranesh2: yea must have been added by the IDE by default and just slipped in. Will change it

deepchem/models/optimizers.py
```diff
@@ -678,3 +677,44 @@ def _create_jax_schedule(self):
...
+
+class KFAC(Optimizer):
+  """The Second order gradient optimiation algorithm which uses an approximation to calculate the inverse of the Fischer matrrix"""
+
+  def __init__(self, **kwargs):
+    """
+    Parameters:
+    -----------
+    model: torch.nn.Module
+    The model to be optimized.
```
peastman, line 690: The description line for each argument needs to be indented.  See https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard.

deepchem/models/tests/test_optimizers.py
```diff
@@ -255,13 +262,70 @@ def test_linearCosine_decay_jax(self):
...
     assert isinstance(jaxopt, optax.GradientTransformation)
+
+  @pytest.mark.torch
+  def test_KFAC(self):
+    """test creating a KFAC optimizer"""
+    import torch
+    import numpy as np
+
+    np.random.seed(123)
+    """linear layers test"""
```
peastman, line 282: Comments inside functions should start with `#`.  Only docstrings are string literals.

deepchem/models/tests/test_optimizers.py
```diff
@@ -255,13 +262,70 @@ def test_linearCosine_decay_jax(self):
...
+    model = dc.models.TorchModel(pytorch_model,
+                                 dc.models.losses.L2Loss(),
+                                 optimizers=optimizers.KFAC(model=pytorch_model,
+                                                            learning_rate=0.1,
+                                                            Tinv=50))
+    model.fit(dataset, nb_epoch=100)
+
+    scores = model.evaluate(dataset, [metric])
+    assert scores[metric.name] > 0.9
+    """Conv2d layers test"""
```
peastman, line 303: Same here.

deepchem/models/torch_models/kfac_optimizer.py
```diff
@@ -0,0 +1,430 @@
...
+      self.m_gg[module] *= self.stat_decay + gg * (1 - self.stat_decay)
+
+  def _prepare_model(self):
+    """"
+    Attaches hooks(saving the ouptut and grad according to the update function) to the model for
+    to calculate gradients at every step.
+    """
+    count = 0
+    print(self.model)
+    print("=> We keep following layers in KFAC. ")
```
peastman, line 257: Logging should be done with the `logging` package rather than `print()` calls.  But is this really intended to be logged, or is it just debugging code that got left in?

peastman:
Mypy seems to think `Tensor.view` is a Tensor rather than a function that returns a Tensor.  I'm not sure why.

I do understand the errors it's reporting in lines 138-9.  You declare `padding` to have type `Union[str, Tuple[int, ...]]`.  That means it could be either a string or a tuple.  But those lines assume it's a tuple of ints.  Assuming `padding` really can be a `str`, those lines are incorrect.  You need to first check `if isinstance(padding, str)` and handle that case differently.  Or if it really has to be a tuple, then you need to fix the declaration.

peastman:
Possibly the problem is that you're circumventing the type system by passing in a layer name as a string.  Like in these lines:

```python
    if classname == 'Conv2d':
      p_grad_mat = m.weight.grad.data.view(m.weight.grad.data.size(0),
                                           -1)  # n_filters * (in_c * kw * kh)
```

You know that if `classname` is `'Conv2d'`, that means `m` is `torch.nn.Conv2d` object, and therefore it has a field called `weight` which is a Tensor.  But mypy doesn't know that.  And of course, you don't really know it either.  That's how you intend the function to be called, but you can't guarantee it always will be.  What if you eliminated the `classname` argument and changed those lines to

```python
    if isinstance(m, torch.nn.Conv2d):
      p_grad_mat = m.weight.grad.data.view(m.weight.grad.data.size(0),
                                           -1)  # n_filters * (in_c * kw * kh)
```

shaipranesh2:
@peastman @tonydavis629 I have done some changes by adding in isinstance to check the type of the weight.grad.data and bias.grad.data and to proceed further only if they are torch tensor type, otherwise raises an error. 
Also, have done the same for the padding , where it checks for str, int and tuple, and proceeds to add the padding accordingly.

deepchem/models/torch_models/kfac_optimizer.py
```diff
@@ -0,0 +1,477 @@
...
+    elif isinstance(padding, str):
+      if padding == 'VALID':
+        pad = int((kernel_size[0] - 1) / 2)
+        x = torch.nn.functional.pad(x, (pad, pad, pad, pad)).data
+
+      x = x.unfold(2, kernel_size[0], stride[0])
+      x = x.unfold(3, kernel_size[1], stride[1])
+      x = x.transpose_(1, 2).transpose_(2, 3).contiguous()
+      x = x.view(x.size(0), x.size(1), x.size(2),
+                 x.size(3) * x.size(4) * x.size(5))
```
peastman, line 164: Are you sure the indentation here is correct?  These lines are only supposed to be executed when `padding` is a string?

peastman (approved):
Assuming all the test failures are unrelated (I haven't checked), this looks good.

shaipranesh2:
@rbharath can you merge this

tonydavis629 (approved):

PR #2994, opened 2022-07-13
update unit tests with DC Lightning dataset module

Chahalprincy:
# Pull Request Template

## Description

This PR modifies the unit tests according to the changes implemented in  #2993

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/lightning/__init__.py
+++ b/deepchem/models/lightning/__init__.py
@@ -1,2 +1,3 @@
 # flake8:noqa
 from deepchem.models.lightning.dc_lightning_module import DCLightningModule
+from deepchem.models.lightning.dc_lightning_dataset_module import DCLightningDatasetModule
--- a/deepchem/models/lightning/dc_lightning_dataset_module.py
+++ b/deepchem/models/lightning/dc_lightning_dataset_module.py
@@ -1,0 +1,36 @@
+import pytorch_lightning as pl
+import torch
+import numpy as np
+
+
+class DCLightningDatasetBatch:
+
+  def __init__(self, batch):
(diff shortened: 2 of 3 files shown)


peastman:
Could you explain what problem DCLightningDatasetModule is meant to solve?  I looked through the documentation on [datamodules](https://pytorch-lightning.readthedocs.io/en/stable/extensions/datamodules.html), and they don't really seem to be relevant to this case.  Here's what it says.

> A datamodule encapsulates the five steps involved in data processing in PyTorch:
> 1. Download / tokenize / process.
> 2. Clean and (maybe) save to disk.
> 3. Load inside [Dataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.Dataset).
> 4. Apply transforms (rotate, tokenize, etc…).
> 5. Wrap inside a [DataLoader](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader).

DCLightningDatasetModule doesn't do any of those, except for the trivial step of wrapping it inside a DataLoader.  And it adds a lot of overhead by forcing the Dataset to provide individual samples rather than complete batches, and then merging them into new numpy arrays.

peastman:
> I can maybe do this without creating the dataset inside the DCLightningDataModule and instead directly returning the dataloader from Deepchem

I don't think it's a question of where the DataLoader is created.  The important point is that you disable automatic batching in the DataLoader and let the Dataset handle it instead.  See https://pytorch.org/docs/stable/data.html#disable-automatic-batching.  That means passing the batch size to `make_pytorch_dataset()`, and specifying `batch_size=None` to the DataLoader.

PR #2997, opened 2022-07-15
Adding electron input streams for Ferminet

shaipranesh2:
# Pull Request Template

## Description
This is a one of several PRs aimed at implementing Ferminet. This PR aims to implement the input stream function
which calculate the input needed for the model (one-electron and two-electron stream).

Fix #(issue)

## Type of change

## Checklist


requirements/env_common.yml
```diff
@@ -26,3 +26,4 @@ dependencies:
     - xgboost
     - git+https://github.com/samoturk/mol2vec
     - tb-nightly # @arunppsg shift to tensorboard stable version once tb 2.9.1 is released
+    - pyscf
```
peastman, line 29: What is this needed for?  Adding a new dependency to DeepChem should be done only when absolutely necessary and needs to be carefully considered first.

shaipranesh2: Yeah I wanted to discuss this over in the developer calls as well. I thought this was necessary as ferminet needs a baseline solution of Hartree-Fock for pre-training which can be obtained from pySCF.

tonydavis629: We can add it as a soft dependency, no? 

I think it may be better to use the pyscf conda channel to install over installing through pip.

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,105 @@
...
+    no_electrons = []
+    nucleons = []
+
+    for i in self.nucleon_coordinates:
+      mol=Chem.MolFromSmiles(i[0])
+      for j in mol.GetAtoms():
+        nucleons.append(j.GetAtomicNum()-j.GetFormalCharge())
+      nucleons.append(i[1])
+
+    electron_no = np.array(no_electrons)
```
tonydavis629, line 72: Why not just assigned electron_no as an empty array?

deepchem/models/tests/test_ferminet.py
```diff
@@ -0,0 +1,37 @@
...
+  h2_molecule = [['[H]', [0, 0, 0]], ['[H]', [0, 0, 0.748]]]
+  molecule = Ferminet(h2_molecule, seed_no=0, batch_number=1)
+  molecule.prepare_input_stream()
+
+  assert (molecule.one_electron_vector == np.array(
+      [[[0.035281046919353284, 0.008003144167344467, 0.019574759682114785],
+        [0.035281046919353284, 0.008003144167344467, -0.7284252403178852]],
+       [[0.044817863984029156, 0.03735115980299935, 0.7284544424024718],
+        [0.044817863984029156, 0.03735115980299935,
+         -0.019545557597528185]]])).all()
```
peastman, line 26: It would be better to use `allclose()` for these assertions.  Very tiny changes, even just running on a different processor, can affect values at this level of precision.

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,127 @@
...
+
+  def __init__(self,
+               nucleon_coordinates: List[List],
+               seed_no: Optional[int] = None,
+               batch_number: int = 10):
+    """
+    Parameters:
+    -----------
+    nucleon_coordinates:  List[List]
+      A dictionary containing nucleon coordinates as the values with the keys as the element's symbol.
```
peastman, line 42: It's a list, not a dict.

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,127 @@
...
+    This method is based on the following paper:
+
+    Spencer, James S., et al. Better, Faster Fermionic Neural Networks. arXiv:2011.07125,
+    arXiv, 13 Nov. 2020. arXiv.org, http://arxiv.org/abs/2011.07125.
+    """
+
+  def __init__(self,
+               nucleon_coordinates: List[List],
+               seed_no: Optional[int] = None,
+               batch_number: int = 10):
```
peastman, line 37: ElectronSampler calls these arguments `seed` and `batch_no`.  We should make the two classes consistent.

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,127 @@
...
+    two_electron_vector: numpy.ndarray
+      The two-electron input stream containing the distance vector between all the electrons coordinates.
+    two_electron_distance: numpy.ndarray
+    """
+
+    no_electrons = []
+    nucleons = []
+
+    for i in self.nucleon_coordinates:
+      mol = Chem.MolFromSmiles(i[0])
```
peastman, line 78: Are they actually SMILES strings, or just element symbols as the docstring says?

shaipranesh2: it should be just the elements symbol. Now, I have made changes such that works with element symbols

deepchem/models/torch_models/ferminet.py
shaipranesh2: I didn't quite understand the "class consistency" here 😅. Did u mean like I should remove the seed_no and batch_number here in ferminet class? If so, then how will the user set the seed and batch number then?

tonydavis629: I think he means the name of the arguments should be changed from `seed_no` to `seed` and `batch_number` to `batch_no`.

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,132 @@
...
+    -----------
+    nucleon_coordinates:  List[List]
+      A list containing nucleon coordinates as the values with the keys as the element's symbol.
+    seed_no: int, optional (default None)
+      Random seed to use for electron initialization.
+    batch_number: int, optional (default 10)
+      Number of batches of the electron's positions to be initialized.
+
+    """
+    super(Ferminet, self).__init__()
```
tonydavis629, line 50: There is no super class for Ferminet to call

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,132 @@
...
+
+    no_electrons = []
+    nucleons = []
+    self.charge: List = []
+
+    for i in self.nucleon_coordinates:
+      mol = Chem.MolFromSmiles('[' + i[0] + ']')
+      for j in mol.GetAtoms():
+        self.charge.append(j.GetAtomicNum())
+        no_electrons.append([j.GetAtomicNum() - j.GetFormalCharge()])
```
peastman, line 84: This is a very expensive way to look up the atomic number for a symbol.  A much cheaper way is

```python
table = Chem.GetPeriodicTable()
...
num = table.GetAtomicNumber(i[0])
```

The formal charge will always be 0, because you don't specify a formal charge.  But should there be an argument somewhere for the user to specify the total charge of the system?

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,132 @@
...
+      nucleons.append(i[1])
+
+    self.electron_no: np.ndarray = np.array(no_electrons)
+    self.nucleon_pos: np.ndarray = np.array(nucleons)
+
+    spin = np.sum(self.electron_no) % 2
```
peastman, line 90: Is this meant to be the total spin of the system, or is it something specific to the model?  Normally the user specifies the total spin, since the ground state isn't always the one with the smallest spin.

shaipranesh2: yeah, this is meant for a single atom, have changed it now to get the user's input

deepchem/models/torch_models/ferminet.py
shaipranesh2: Have changed it now. and for specifying the formal charge, I have allowed the user to input it with the element string itself, like 'H1+' or 'H1-', where the end of the string defines the charge type and the number before it defines the magnitude. Is this fine?

peastman: Usually the way quantum chemistry programs handle this is that you specify the elements, the total charge, and the spin multiplicity.  See for example https://psicode.org/psi4manual/master/tutorial.html.  By specifying formal charges, you imply that you know where all the electrons are going to end up.  But generally you don't know, and it's the program's job to figure out where they'll go.

shaipranesh2: Since at the start we sample electrons around each atom, in case of a formal charge, an atom is randomly selected for the charge?(I am unsure on how the program will figure out at the start)

shaipranesh2:
@peastman @tonydavis629 , for the electronegativity data, I found this csv (https://github.com/lewisacidic/scikit-chem/blob/cbc9765e63cc78e771d6c31b5c708dfaf2a19df5/skchem/resource/atom_data.csv), containing electronegative and a few other properties. So should I go ahead with this data?

peastman:
Sure, it seems worth a try.  Wikipedia also has [several different electronegativity scales](https://en.wikipedia.org/wiki/Electronegativities_of_the_elements_(data_page)).  I don't think it matters which one we use.  We just need some reasonable way of ordering the elements.

shaipranesh2:
@tonydavis629 @peastman i guess the CI errors are not related to this code. But is this way of calculating the electronegativity fine?

shaipranesh2:
@tonydavis629 @peastman the errors are not related to the code here I guess. And is this way of using electronegativity fine?

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,202 @@
...
+      seed: Optional[int] = None,
+      batch_no: int = 10,
+  ):
+    """
+    Parameters:
+    -----------
+    nucleon_coordinates:  List[List]
+      A list containing nucleon coordinates as the values with the keys as the element's symbol.
+    spin: int
+      The total spin of the molecule system.
```
peastman, line 56: You probably mean spin multiplicity?  Spin isn't an integer.

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,202 @@
...
+      elif len(
+          nucleons
+      ) == 2:  # for a diatomic molecule, the most electronegative atom will get the anionic charge
+        electro_neg = electro_neg[electro_neg[:, 1].argsort()]
+        if self.ion_charge > 0:
+          pos = electro_neg[0][0]
+        else:
+          pos = electro_neg[-1][0]
+        self.electron_no[int(pos)][0] -= self.ion_charge
+      else:  # for a multiatomic molecule, the atom's electronegativity is averaged out with the weight sum of neighbouring atom's electronegativity and their interatomic distance
```
peastman, line 125: I'm not familiar with this calculation.  But I'm also not a chemist.  What's the reason for it?

deepchem/models/torch_models/ferminet.py
```diff
@@ -0,0 +1,202 @@
...
+            self.electron_no[int(electro_neg[pos][0])][0] -= per_atom_charge
+            pos += increment
+          self.electron_no[int(electro_neg[pos -
+                                           increment][0])][0] += extra_charge
+        else:
+          if self.ion_charge > 0:
+            pos = 0
+          else:
+            pos = -1
+          self.electron_no[int(electro_neg[pos][0])][0] -= self.ion_charge
```
peastman, line 157: Adding multiple extra electrons to a single atom probably won't work well.  Highly charged atoms are unusual and tend to have high energies.  For example, if the total charge is -2 you probably should assign -1 to each of the two most electronegative atoms, not -2 to a single atom.

deepchem/models/torch_models/ferminet.py
peastman: Remember to remove the comment, since it doesn't apply anymore.

deepchem/models/torch_models/ferminet.py
peastman: Remember to fix this.

peastman (approved):

Issue #3006, opened 2022-07-25
Converting large disk dataset to csv


arunppsg:
When there is a requirement to convert disk dataset to csv, currently we can only do it be converting it to a dataframe and then to a csv. But a drawback of this approach is that DiskDatasets which are large in size generate dataframes which don't fit in memory.  Having a method which can convert a large diskdataset to csv without consuming memory will be useful.

I would like to suggest an improvement to `class Dataset` by adding a `to_csv` method. This method can then use `itershards` to write data to a csv file.

peastman:
`itershards()` is a method of DiskDataset.  If we want to put the method in the parent Dataset class (which seems like the best place for it, so you can write out any dataset), we can use `itersamples()`.

arunppsg:
Will using `iterbatches()` be good? Because when I use `iterbatches()` or `itershards()`, I can read a batch of data and write to a csv file in chunk. But when I use `itersamples()`, I have to either perform a write operation for each sample, or I need to use a separate variable to create a buffer for storing data temporarily.

peastman:
`iterbatches()` can work too.  In your PR I see you're using Pandas to construct DataFrames from blocks of samples and append them to the file.  I had assumed you would use the standard `csv` module, which writes rows one at a time.  But Pandas can work too.

arunppsg:
It makes more sense to use standard `csv` module since it will use less memory. I have updated the PR to use that approach.

PR #3007, opened 2022-07-25
Adding to_csv method to DiskDataset

arunppsg:
Fix for #3006 

In this PR, I have added a `to_csv` method which creates a dataframe from a `Dataset` object via the following steps:
1. Iterate through the shards
2. Create dataframe for each shard
3. Append the dataframe to the csv file.

The method accepts path to the csv file as an argument and it returns None.

### DIFF
--- a/deepchem/data/datasets.py
+++ b/deepchem/data/datasets.py
@@ -3,6 +3,7 @@ Contains wrapper class for datasets.
 """
 import json
 import os
+import csv
 import math
 import random
 import logging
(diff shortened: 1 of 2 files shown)


arunppsg:
@rbharath I considered having `to_csv` as a staticmethod but it did not make much sense to me. The reason for not having is staticmethod is because `to_csv` operates on the instance's data and not on any static properties of the class.

peastman:
It looks like this assumes the `x` and `y` values for each sample are either scalars or 1D arrays.  What happens if they're higher dimensional?  For example, what if `x` is a 2D image for each sample?  And more important, what *should* happen in that case?

Similarly, what if they're objects such as GraphData?  Should we try to write them out in some format that can be reconstructed again?  Or throw an exception saying they can't be written to csv?

arunppsg:
I think to begin with, we can fail loudly - check whether the input data is writable, if not raise an assertion error that `to_csv` method is only valid for cases were X is a scalar or a 1-D array.

peastman:
That sounds good.

rbharath (approved):
LGTM! Will leave open in case @peastman wants to review

peastman:
Looks good to me.

PR #3035, opened 2022-08-19
Multigpu

Chahalprincy:
Make the loss implementation non-local, if the implementation is local then when launching processes for multiple GPUs code breaks. Python uses pickle in the process of forking processes and the pickle fails with the error `torch multigpu AttributeError: Can't pickle local object`.

Questions:
1. Is there a better way to do this, one possibility is with https://docs.python.org/3/reference/expressions.html#lambda, I will have to test this way more to make things work.
2. This would require making changes to how loss is specified in other methods also. 

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/losses.py
+++ b/deepchem/models/losses.py
@@ -63,10 +63,17 @@ class HuberLoss(Loss):
 
     def loss(output, labels):
       output, labels = _make_pytorch_shapes_consistent(output, labels)
-      return torch.nn.functional.smooth_l1_loss(
-          output, labels, reduction='none')
+      return torch.nn.functional.smooth_l1_loss(output,
+                                                labels,
+                                                reduction='none')
 
     return loss
+
+
+def l2_loss(output, labels):
+  import torch
+  output, labels = _make_pytorch_shapes_consistent(output, labels)
+  return torch.nn.functional.mse_loss(output, labels, reduction='none')
 
 
 class L2Loss(Loss):
@@ -79,13 +86,7 @@ class L2Loss(Loss):
     return tf.square(output - labels)
 
   def _create_pytorch_loss(self):
-    import torch
-
-    def loss(output, labels):
-      output, labels = _make_pytorch_shapes_consistent(output, labels)
-      return torch.nn.functional.mse_loss(output, labels, reduction='none')
-
-    return loss
+    return l2_loss
 
 
 class HingeLoss(Loss):
@@ -112,7 +113,7 @@ class HingeLoss(Loss):
 
 class SquaredHingeLoss(Loss):
   """The Squared Hinge loss function.
-  
+
   Defined as the square of the hinge loss between y_true and y_pred. The Squared Hinge Loss is differentiable.
   """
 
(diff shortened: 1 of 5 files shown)


deepchem/models/losses.py
```diff
@@ -63,12 +63,19 @@ def _create_pytorch_loss(self):
...
-      return torch.nn.functional.smooth_l1_loss(
-          output, labels, reduction='none')
+      return torch.nn.functional.smooth_l1_loss(output,
+                                                labels,
+                                                reduction='none')
 
     return loss
 
 
+def l2_loss(output, labels):
```
peastman, line 73: I think moving the definition out of the function is fine, if that avoids problems with lightning.  Let's rename it to `_pytorch_l2_loss()` to clarify that it's private and specific to PyTorch.

deepchem/data/datasets.py
```diff
@@ -1607,10 +1607,18 @@ def iterate(dataset: DiskDataset, batch_size: Optional[int], epochs: int):
       # mp.dummy aliases ThreadPool to Pool
       pool = Pool(1)
 
+      import torch.distributed as dist
```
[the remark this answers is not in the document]
peastman: We can't import `torch` here.  This method isn't specific to pytorch.

deepchem/data/datasets.py
```diff
@@ -1607,10 +1607,18 @@ def iterate(dataset: DiskDataset, batch_size: Optional[int], epochs: int):
...
+        world_size = dist.get_world_size()
+      else:
+        world_size = 1
+
       if batch_size is None:
         num_global_batches = num_shards
       else:
-        num_global_batches = math.ceil(dataset.get_shape()[0][0] / batch_size)
+        # num_global_batches = math.ceil(dataset.get_shape()[0][0] / batch_size)
+        num_global_batches = (dataset.get_shape()[0][0] // world_size) // batch_size
```
peastman, line 1621: Why did you remove the `math.ceil()` and change `/` to `//`?  When you have a final partial batch, that will cause it to be skipped.

peastman (commented):
Can you explain what the issues are?  The existing code (calling `get_worker_info()`) already supports multiple processes (not just threads).  What are the situations where it doesn't work?

peastman:
I don't think you should need any changes at all to `_iterbatches_from_shards()`.  Each shard will still be handled by a single process, and it's up to `__iter__()` to assign them.  The terms `local` and `global` in that method aren't about processes.  They refer to a single shard, and to the full list of shards that was passed in `shard_indices`.

I think the changes to `__iter__()` are more complicated than they need to be.  It looks to me like something along these lines would work?

```python
if worker_info is None:
    process_id = 0
    num_processes = 1
else:
    process_id = worker_info.id
    num_processes = worker_info.num_workers
if dist.is_initialized():
    process_id += dist.get_rank() * num_processes
    num_processes *= dist.get_world_size()
first_shard = process_id * n_shards // num_processes
last_shard = (process_id + 1) * n_shards // num_processes
```

PR #3038, opened 2022-08-23
Fixes to k-fold fingerprint splitting

chertianser:
# Pull Request Template

## Description

There are two issues when attempting to k-fold split via fingerprint similarity:

1.  If a dataset of size `N` does not divide evenly into `k` folds (i.e `N mod k != 0`) the extra data points are discarded. As `test_size != 0` when `N mod k != 0`, there is an attempted second split for the test indices. `k_fold_split` for `FingerprintSplitter` only collects training and validation data, and therefore the test indices are lost. The proposed fix by adding checks on `frac_valid` and `frac_test` ensure that the extra data points are merged appropriately into the test and validation sets respectively.
2. There is a division-by-zero error when attempting to k-fold split by fingerprint similarity. At the final `k`th fold, `split` calls with arguments `frac_train=1.0`, `frac_valid=0.0` and `frac_test=0.0`. `_split_fingerprints` is then called with `size1=N/k` and `size2=0`, which will fail as the grouping algorithm attempts to divide by zero. A straightforward fix is to return the identity (i.e all of the unsplit indices), because the final fold is already considered split by process of elimination.

To test on both these issues, I have also written a unit test that checks for both issues.

Extra line changes were caused by running `yapf` on both `splitter.py` and `test_splitter.py`.

## Type of change

## Checklist

### DIFF
--- a/deepchem/splits/splitters.py
+++ b/deepchem/splits/splitters.py
@@ -83,12 +83,11 @@ class Splitter(object):
       # to k-1.
       frac_fold = 1. / (k - fold)
       train_dir, cv_dir = directories[2 * fold], directories[2 * fold + 1]
-      fold_inds, rem_inds, _ = self.split(
-          rem_dataset,
-          frac_train=frac_fold,
-          frac_valid=1 - frac_fold,
-          frac_test=0,
-          **kwargs)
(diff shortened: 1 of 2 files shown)


arunppsg:
Looks great. Thank you @chertianser

arunppsg (approved):

PR #3047, opened 2022-08-28
Assigning task names when they are not specified

### LINKED ISSUE #3044: Shape of DiskDataset attributes is not getting displayed when task names are not specified
Sample code to reproduce:
```
def foo():
    for i in range(1):
        yield np.random.randn(10, 1), np.random.randn(10, 1), np.ones(10), np.arange(10)

output_dataset = dc.data.DiskDataset.create_dataset(shard_generator=foo())

output_dataset.get_shape()
```
The above line prints `((10, 1), (), (), (10,))` but expected `((10, 1), (10, 1), (10,), (10,))`.  Similarly, `output_dataset.get_task_names()` prints `array([], dtype=float64)`.

Since `DiskDataset.create_dataset` assigns an empty list for task names as a default value and `get_shape` depends on `get_task_names` to compute y shape, it is throwing this error.

arunppsg:
Fix #3044

I added a small check condition to look for tasks field length. It it is zero, set task names as 0, 1, 2, ... based on `y.shape` during dataset creation.


rbharath (commented):
Looks good on my end. Once CI is clean we can merge in

PR #3051, opened 2022-09-02
pinning scipy to fix torch build in ci

arunppsg:


chajath:
Why was this necessary? SciPy is widely used so this isn't gonna scale well

arunppsg:
We had a build failure in CI tests for pytorch when using scipy version>=1.9. In particular, [this](https://github.com/deepchem/deepchem/actions/runs/2973337100/jobs/4762830402#step:6:22) was the failure. We observed it only in environment with pytorch not in other envs. But nevertheless, we should unpin it soon.

PR #3065, opened 2022-09-15
Correctly import lightning to avoid import errors

abhiskk:
## Description

Fix deepchem installation failure, remove try catch block around lightning import which was causing previous crashes.

## Type of change

Tested the fix with
```
pip install --pre deepchem
import deepchem as dc
print(dc.__version__)
```


rbharath (approved):
LGTM!

Issue #3098, opened 2022-10-31
MultiTaskClassifier is not working


andresilvapimentel:
## To Reproduce

The piece of code below in the tutorial Introduction_to_Model_Interpretability.ipynb is not working

n_tasks = len(tasks)
n_features = train_dataset.get_data_shape()[0]
model = dc.models.MultitaskClassifier(n_tasks, n_features)
model.fit(train_dataset, nb_epoch=50)

The error message is:  AttributeError: module 'deepchem.models' has no attribute 'MultitaskClassifier

## Environment

* OS: Windows
* Python version: 3.7.15
* DeepChem version: 2.6.2.dev
* RDKit version (optional):
* TensorFlow version (optional):
* PyTorch version (optional):
* Any other relevant information: on Google colab

arunppsg:
MultitaksClassifier requires pytorch. Could you check whether you have it in your environment?

andresilvapimentel:
Yes, it is.
I solved the issue installing the version 2.6.1, but there is an issue with the version 2.6.2.

arunppsg:
I think the import fails when there is no torch-geometric package even though MultitaskClassifier does not depend on torch-geometric. I will able to reproduce it and I will put a fix soon.

PR #3105, opened 2022-11-11
Minor patches

arunppsg:
This PR fixes:
- docs ci by updating torch-geometric import
- making models requiring torch-geometric import under an separate try-except block, so that torch-geometric import fail does not block import of other models (potential fix for #3098)
- updates torch version to 1.12 (we have it pinned because torch-geometric requires a precise version of pytorch)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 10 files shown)


arunppsg:
Torch CI is failing with a seg-fault.

rbharath (commented):
Did a review, had some minor style comments for @arunppsg offline about cleanup but good to merge in once cleaned up

PR #3135, opened 2022-12-12
Updating RDKitDescriptors featurizer to support custom descriptors

gusty1g:
This PR updates RDKitDescriptors to support user-defined descriptors. For this, an additional arguments, `descriptors` which takes in a `List[str]` is added. It takes a list of custom user defined descriptors and produce output properties.

An another change is adding `labels_only` argument which when set to True, it returns only the presence or absence of a group.

New tests were also added to cover the new features.

### DIFF
--- a/deepchem/feat/molecule_featurizers/rdkit_descriptors.py
+++ b/deepchem/feat/molecule_featurizers/rdkit_descriptors.py
@@ -1,8 +1,7 @@
 """
 Basic molecular features.
 """
-
-from typing import List, Sequence, Tuple, Union, Callable, Dict
+from typing import List, Sequence, Union, Callable, Dict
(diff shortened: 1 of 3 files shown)


gusty1k:
In 6364020, I removed the test assertions using array index for molecular weight. The array index for molecular weight can be any value based on the sorted descriptors list. A descriptor list with only two descriptors, say [ExactMolWt, BCUT2D_MWHI]  will have ExactMolWt at index 1 (sorted) and for the descriptor list [ExactMolWt], ExactMolWt will be at index 0. Since the array index is not deterministic, I removed it from test assertion.

gusty1g:
Reviewed by @tonydavis629 offline.

Issue #3154, opened 2022-12-30
Refactoring dmpnn featurizer


gusty1g:
I have a couple of suggestions to refactor dmpnn featurizer (`molecule_featurizers/dmpnn_featurizer.py`):
- The class _MapperDMPNN is not being used. It is also a duplicate of `dc.models.torch_models.dmpnn._MapperDMPNN` class. Probably, `dmpnn_featurizer._MapperDMPNN` can be safely removed.
- The methods `atom_features` and `bond_features` are unique to DMPNNFeaturizer and they should be a classmethod.
- Some functions like `get_atom_chiral_tag_one_hot` contains only one line of code which invokes another function (`one_hot_encode`). These functions can be removed and the inner function can be directly used. Ex:
https://github.com/deepchem/deepchem/blob/d205e2dd0c3671fbae00faee1ee946c6937575c2/deepchem/feat/molecule_featurizers/dmpnn_featurizer.py#L108

The featurizer and ideas from DMPNN paper are also used in other message passing networks like Grover. It will be good to clean up DMPNN and standardize it, so that code duplication can be prevented.

PR #3158, opened 2023-01-03
Refactoring DMPNN: removing _MapperDMPNN class

gusty1g:
Partial fix for #3154 .

In this PR, I removed _MapperDMPNN since it is not used.


rbharath (approved):
LGTM

PR #3179, opened 2023-01-16
Adding grover layers

gusty1g:
I have added layers which are used by Grover Model in this PR. The layers are:
- GroverMPNEncoder - the layer performs message passing
- GroverAttentionHead - uses GroverMPNEncoder as encode to generate query, key and value for attention network
- GroverMTBlock - A wrapper layer which wraps GroverMPNEncoder and GroverAttentionHead. It drives message passing with attention layer
- GroverTransEncoder
- GroverEmbedding

### DIFF
--- a/deepchem/models/torch_models/grover.py
+++ b/deepchem/models/torch_models/grover.py
@@ -4,12 +4,39 @@ from typing import List
 from deepchem.models.torch_models.grover_layers import GroverTransEncoder
 
 
-class GROVEREmbedding(nn.Module):
+class GroverEmbedding(nn.Module):
+    """GroverEmbedding layer.
+
+    This layer is a simple wrapper over GroverTransEncoder layer for retrieving the embeddings from the GroverTransEncoder corresponding to the `atom_embedding_output_type` chosen by the user.
+
+    Parameters
+    ----------
+    edge_fdim: int
+        the dimension of additional feature for edge/bond.
+    node_fdim: int
+        the dimension of additional feature for node/atom.
+    depth: int
+        Dynamic message passing depth for use in MPNEncoder
+    undirected: bool
+        The message passing is undirected or not
+    num_mt_block: int
+        the number of message passing blocks.
+    num_head: int
+        the number of attention heads.
+    atom_embedding_output_type: str
+        the type of output aggregation after message passing.
+                                            atom_messages:      True                      False
+        - "none": no aggregation         output size:     (num_atoms, hidden_size)    (num_bonds, hidden_size)
+        -"atom":  aggregating to atom  output size:     (num_atoms, hidden_size)    (num_atoms, hidden_size)
+        -"bond": aggragating to bond.   output size:     (num_bonds, hidden_size)    (num_bonds, hidden_size)
+        -"both": aggregating to atom&bond. output size:  (num_atoms, hidden_size)    (num_bonds, hidden_size)
+                                                        (num_bonds, hidden_size)    (num_atoms, hidden_size)
+    """
 
     def __init__(self,
-                 node_dim,
-                 edge_dim,
-                 embedding_output_type,
+                 node_fdim,
+                 edge_fdim,
+                 atom_embedding_output_type,
                  hidden_size=128,
                  depth=1,
                  undirected=False,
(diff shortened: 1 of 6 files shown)


deepchem/models/torch_models/grover_layers.py
```diff
@@ -0,0 +1,688 @@
...
+        if self.input_layer == 'fc':
+            message = self.act_func(self.W_i(init_messages))
+        elif self.input_layer == 'none':
+            message = init_messages
+
+        attached_feats = init_attached_features
+
+        if self.training and self.dynamic_depth != 'none':
+            if self.dynamic_depth == 'uniform':
+                ndepth = int(np.random.uniform(self.depth - 3, self.depth + 3))
```
gusty1g, line 161: FIXME: If depth is less than 3, ndepth can be negative. This should be fixed.

deepchem/models/torch_models/grover.py
```diff
@@ -0,0 +1,67 @@
+import torch
+import torch.nn as nn
+from typing import List
+from deepchem.models.torch_models.grover_layers import GroverTransEncoder
+
+
+class GROVEREmbedding(nn.Module):
+
```
tonydavis629, line 8: This class needs a doc string

deepchem/models/torch_models/grover_layers.py
```diff
@@ -0,0 +1,688 @@
...
+            raise ValueError('Only ReLU activation function is supported')
+
+        if self.input_layer == "fc":
+            input_dim = self.init_message_dim
+            self.W_i = nn.Linear(input_dim, hidden_size, bias=bias)
+
+        # Bug here, if input_layer is none, then w_h_input_size should have init_messages_dim
+        # instead of hidden_size for first iteration of message passing alone (in subsequent
+        # iterations, message size will be n_bonds x hidden_size but during first iteration,
+        # message size will be init_messages_dim
```
tonydavis629, line 135: Maybe raise an error in this case?

deepchem/models/torch_models/grover_layers.py
```diff
@@ -0,0 +1,688 @@
...
+                 bias: bool = True,
+                 depth: int = 1,
+                 dropout: float = 0.0,
+                 undirected: bool = False,
+                 atom_messages: bool = False):
+        super(GroverAttentionHead, self).__init__()
+        self.atom_messages = atom_messages
+
+        # FIXME We assume that we are using a hidden layer to transform the initial atom message
+        # and bond messages to hidden dimension size.
```
tonydavis629, line 230: This expectation should probably be in the doc string

deepchem/models/torch_models/tests/test_grover_layers.py
```diff
@@ -0,0 +1,143 @@
...
+    f_atoms = torch.randn(4, 151)
+    f_bonds = torch.randn(5, 165)
+    a2b = torch.Tensor([[0, 0], [2, 0], [4, 0], [1, 3]]).type(torch.int32)
+    b2a = torch.Tensor([0, 1, 3, 2, 3]).type(torch.long)
+    b2revb = torch.Tensor([0, 2, 1, 4, 3]).type(torch.long)
+    a2a = torch.Tensor([[0, 0], [3, 0], [3, 0], [1, 2]]).type(torch.int32)
+
+    # TODO Write tests for undirected = True case, currently fails. for this case, we have
+    # to generate inputs (a2b, b2a, b2revb) for undirected graph (mol2graph returns features
+    # for directed graphs)
```
tonydavis629, line 23: What is mol2graph?

deepchem/models/torch_models/tests/test_grover_layers.py
```diff
@@ -0,0 +1,143 @@
...
+        [f_atoms, f_bonds, a2b, b2a, b2revb, a_scope, b_scope, a2a])
+    new_f_atoms, new_f_bonds, new_a2b, new_b2a, new_b2revb, new_a_scope, new_b_scope, new_a2a = new_batch
+    assert new_f_atoms.shape == (f_atoms.shape[0], hidden_size)
+    assert new_f_bonds.shape == f_bonds.shape
+    assert (new_a2b == a2b).all()
+    assert (new_b2a == b2a).all()
+    assert (new_b2revb == b2revb).all()
+    assert (new_a_scope == a_scope).all()
+    assert (new_b_scope == b_scope).all()
+    assert (new_a2a == a2a).all()
```
tonydavis629, line 119: Can you add a comment explaining why we expect the MTBlock to return the same input values?

tonydavis629 (changes requested):
Code looks good, just needs a little more explanation in the documentation

deepchem/models/torch_models/tests/test_grover_layers.py
gusty1g: That was a note to self, based on original implementation of grover. I have removed it since it is not relevent here anymore.

deepchem/models/torch_models/tests/test_grover_layers.py
gusty1g: sure, added it.

tonydavis629 (approved):

PR #3181, opened 2023-01-17
type fixes for pep 484 compliance

gusty1g:
Made some type fixes for compliance with pep 484.

PEP 484 prohibits the use of implicit optional. For example, the following piece of code allows assumption of implicit optional: `def foo(a: int = None): ... `. It allows type checkers to assume an optional type of `None` for `a`. To comply with PEP 484, it should be `def foo(a: Optional[int] = None): ...`.

The latest mypy version (0.991, release [blog](https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html)) had a breaking change which disabled implicit optionals type. Hence, it was not detected earlier by mypy.

### DIFF
--- a/deepchem/data/pytorch_datasets.py
+++ b/deepchem/data/pytorch_datasets.py
@@ -2,6 +2,7 @@ import numpy as np
 import torch
 
 from deepchem.data.datasets import NumpyDataset, DiskDataset, ImageDataset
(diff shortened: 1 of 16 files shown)


rbharath (approved):
LGTM, reviewed at OH

PR #3183, opened 2023-01-17
Adding attention layers

gusty1g:
I am adding support for two attention layers - ScaledDotProductAttention and Self-Attention which will be used in grover model implementation in this PR.


tonydavis629 (approved):
LGTM. Reviewed in code review

PR #3235, opened 2023-02-09
Fix flake8 errors

### LINKED ISSUE #3182: Flake8 checks on deepchem/models directory
I was having a look at `scripts/flake8_for_ci.sh` script which runs flake8 in CI. I noticed that the flake8 tests are currently not run in `deepchem/models` directory. We should add `deepchem/models` to the list of items in `scripts/flake8_for_ci.sh`.

Subsequently, we should also fix the breakages caused by adding flake8 checks for `deepchem/models` directory.

brianpzaide:
# Pull Request Template

## Description

Fix #3182

fixing flake8 error in "deepchem/models dir"

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/IRV.py
+++ b/deepchem/models/IRV.py
@@ -1,10 +1,8 @@
-import logging
 import numpy as np
 import tensorflow as tf
 
 from deepchem.models import KerasModel, layers
 from deepchem.models.losses import SigmoidCrossEntropy
-from deepchem.trans import undo_transforms
 from tensorflow.keras.layers import Input, Layer, Activation, Concatenate, Lambda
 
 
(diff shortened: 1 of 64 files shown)


gusty1g:
Hey, there seems to be some failure in the CI. Some are yapf [errors](https://github.com/deepchem/deepchem/actions/runs/4136999931/jobs/7151611215#step:13:123). The failure on jax CI occurring  as an import error in `test_optimizers.py` seems to be concerning. Any idea why it might be happening?

brianpzaide:
I do not know why there is an import error.
I ran the test locally(I commented out the actual tests)this is what i did
code
here's the snapshot of the result
test
i did not get any import errors.
i am reading on python import system, namespaces may be if I can find something

gusty1g:
@brianpzaide I am not sure 66245c7 will fix the issue.

Summary of change: I replaced the usage of TorchModel with `dc.data.TorchModel`. 

The reason why I am hoping this will fix the issue is in deepchem CI tests for jax, we don't install torch dependencies (pytorch and others) in the workflow. When `from deepchem.models import TorchModel` is at the top, the import will fail because pytorch is not installed.

brianpzaide:
@gusty1g I see, I had not read the github workflows directory earlier, my bad
can I make the changes and resend the pr?
thanks

gusty1g (approved):

PR #3288, opened 2023-03-20
adding nnlda layers

advikavs:
## Description

PR on nnlda layer


advikavs:
re-opened on accident

PR #3302, opened 2023-03-23
save restore for modularTorchModel

gusty1g:
A draft PR demonstrating for saving and restoring ModularTorchModel.


tonydavis629 (commented):
This all makes sense. We will also need to overwrite load_from_pretrained to accept the components argument, and to remove load_pretrained_components.

PR #3315, opened 2023-03-29
fixing doc test - dqc

advikavs:
## Description

fixing doc test - dqc

### DIFF
--- a/deepchem/feat/dft_data.py
+++ b/deepchem/feat/dft_data.py
@@ -17,6 +17,7 @@ try:
     has_dqc = True
 except ModuleNotFoundError:
     has_dqc = False
+    pass
 
 
 class DFTSystem():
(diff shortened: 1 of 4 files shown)


deepchem/feat/dft_data.py
```diff
@@ -21,9 +21,9 @@ class DFTSystem():
 
     Examples
     --------
-    >>> from deepchem.feat.dft_data import DFTSystem
```
advikavs, line 24: do not make >> instead of >>>

rbharath (commented):
Documenting offline discussion, this may be caused by the tensorflow-probability issue @ARY2260 mentioned. Let's try pinning the tf-probability version and if that doesn't fix it, let's discuss again tomorrow

PR #3770, opened 2024-01-05
[WIP] Ray multi-cpu featurization

arunppsg:
An example of ray multi-cpu featurization prototype (wip).


rbharath (commented):
This is really interesting. Can you add some more docs and explain the logic in comments @arunppsg ? I'd be interested in reading through to learn more about ray usage.

CC @vsaravind01

PR #4006, opened 2024-06-09
add: remaining modes to torch.compile

gauthamk02:
## Description

Adding support for the rest of the modes of `torch.compile`

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/tests/test_torch_model.py
+++ b/deepchem/models/tests/test_torch_model.py
@@ -517,30 +517,36 @@ def test_torch_compile():
                           size=(n_samples, n_tasks)).astype(np.float32)
     dataset = dc.data.NumpyDataset(X, y)
 
+    def test_config_overfit(mode='default', backend='inductor'):
+        model = dc.models.torch_models.CNN(n_tasks,
+                                           n_features,
+                                           dims=1,
+                                           kernel_size=3,
+                                           mode='classification')
+        model.compile(mode=mode, backend=backend)
+        model.fit(dataset, nb_epoch=500)
+
+        model_output = model.predict_on_batch(X)
+        model_output = np.argmax(model_output, axis=2)
+        return model_output
+
     # Testing backends
-    inductor_model = dc.models.torch_models.CNN(n_tasks,
-                                                n_features,
-                                                dims=1,
-                                                kernel_size=3,
-                                                mode='classification')
-    inductor_model.compile(backend='inductor')
-    inductor_model.fit(dataset, nb_epoch=500)
-
-    inductor_model_output = inductor_model.predict_on_batch(X)
-    inductor_model_output = np.argmax(inductor_model_output, axis=2)
-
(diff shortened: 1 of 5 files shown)


gauthamk02:
@arunppsg Can you please review the PR.

requirements/env_common.yml
```diff
@@ -27,3 +27,4 @@ dependencies:
     - xgboost
     - gensim  # used by mol2vec
     - tensorboard
+    - triton
```
arunppsg, line 30: Is triton a must requirement?

deepchem/models/tests/test_torch_model.py
```diff
@@ -517,30 +517,36 @@ def test_torch_compile():
...
-                                                n_features,
-                                                dims=1,
-                                                kernel_size=3,
-                                                mode='classification')
-    inductor_model.compile(backend='inductor')
-    inductor_model.fit(dataset, nb_epoch=500)
-
-    inductor_model_output = inductor_model.predict_on_batch(X)
-    inductor_model_output = np.argmax(inductor_model_output, axis=2)
+    def test_config_overfit(mode='default', backend='inductor'):
```
arunppsg, line 520: A test within a test is not a good pattern. An unit-test must test for only one condition at a time. Maybe you can use pytest parameterize.

requirements/env_common.yml
arunppsg: Since the model compilation is done by PyTorch, probably, we can skip it in dependency and test only for one or two modes - thus testing only the API. If the user wants it, they can have it installed.

arunppsg (changes requested):

requirements/env_common.yml
gauthamk02: triton is needed for testing the modes `reduce-overhead` and `max-autotune`. We can remove the test for these two modes and only do testing for `default` and `max-autotune-no-cudagraphs` which doesn't require any additional dependency. Shall I remove the tests for `reduce-overhead` and `max-autotune`?

deepchem/models/tests/test_torch_model.py
gauthamk02: If we remove the tests for other modes and test only the API of `torch.compile` instead, we can remove the subfunction and the test will only check for one condition. So I'll change the test according to how I'm resolving the previous comment.

requirements/env_common.yml
arunppsg: yes

arunppsg (commented):
LGTM

arunppsg (approved):

arunppsg:
Once @rbharath gives a second-pair of eyes, we can merge it in - the CI failures are also not related to this PR.

rbharath (commented):
LGTM to merge in after rebasing (following numpy <2 patch)

rbharath (approved):
LGTM

PR #4508, opened 2025-08-08
lightning trainer

bhuvanmdev:
## Description
This pull request introduces the new `LightningTorchModel` class, providing a high-level interface for training, evaluating, and managing DeepChem models using PyTorch Lightning. It also adds comprehensive tests to verify correct checkpointing and model reload behavior. The changes improve multi-GPU support, reproducibility, and the overall integration of DeepChem models with Lightning.

**Trainer Class:**

* Added the `LightningTorchModel` class in `deepchem/models/lightning/trainer.py`, which wraps DeepChem `TorchModel` instances for training, prediction, evaluation, and checkpointing using PyTorch Lightning. This includes methods for fitting, predicting, evaluating, saving, and loading checkpoints.
* Exposed `LightningTorchModel` in the Lightning module's public API by importing it in `deepchem/models/lightning/__init__.py`.

**Testing and Validation:**

* Added new tests in `deepchem/models/lightning/tests/test_lightning_trainer.py` for checkpoint save/load correctness and model reproducibility for both multitask classifiers and GCN models. These tests ensure weights are preserved across checkpoint reloads and verify that models can overfit small datasets using Lightning.

## Test case results
image

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/lightning/tests/test_lightning_trainer.py
+++ b/deepchem/models/lightning/tests/test_lightning_trainer.py
@@ -2,6 +2,7 @@ import pytest
 import deepchem as dc
 import numpy as np
 import os
+import shutil
 try:
     import torch
     gpu_available = torch.cuda.is_available() and torch.cuda.device_count() > 0
@@ -24,7 +25,70 @@ pytestmark = [
 
 
 @pytest.mark.torch
-def test_multitask_classifier_reload_correctness():
+def test_manual_save_restore():
+    L.seed_everything(42)
+    tasks, datasets, _ = dc.molnet.load_clintox()
+    _, valid_dataset, _ = datasets
+
+    # Create first model and trainer
+    model1 = dc.models.MultitaskClassifier(n_tasks=len(tasks),
+                                           n_features=1024,
+                                           layer_sizes=[1000],
+                                           dropouts=0.2,
+                                           learning_rate=0.0001,
+                                           device="cpu",
+                                           batch_size=16)
+
+    # Use a specific model_dir for this test to avoid conflicts
+    trainer1 = LightningTorchModel(
+        model=model1,
+        batch_size=16,
+        accelerator="cuda",
+        devices=-1,
+        log_every_n_steps=1,
+    )
+
+    # Train first model
+    trainer1.fit(valid_dataset, nb_epoch=3,
+                 checkpoint_interval=0)  # disable auto checkpointing
+
+    trainer1.save_checkpoint(model_dir="test_dir")
+
+    y1 = trainer1.predict(valid_dataset)
+
+    # Create second model and trainer with same model_dir
+    model2 = dc.models.MultitaskClassifier(n_tasks=len(tasks),
+                                           n_features=1024,
+                                           layer_sizes=[1000],
+                                           dropouts=0.2,
+                                           learning_rate=0.0001,
+                                           device="cpu",
+                                           batch_size=16)
+
+    trainer2 = LightningTorchModel(model=model2,
+                                   batch_size=16,
+                                   model_dir="test_dir",
+                                   accelerator="cuda",
+                                   devices=-1,
+                                   log_every_n_steps=1)
+
+    # Restore from specific checkpoint name
+    trainer2.restore()
+
+    # Now they should produce similar results
+    y2 = trainer2.predict(valid_dataset)
+    assert np.allclose(y1, y2, atol=1e-3)
+
+    # Clean up
+    try:
+        if os.path.exists("test_dir"):
+            shutil.rmtree("test_dir")
+    except:
+        pass  # Ignore cleanup errors
+
+
+@pytest.mark.torch
+def test_multitask_classifier_restore_correctness():
     L.seed_everything(42)
     tasks, datasets, _ = dc.molnet.load_clintox()
     _, valid_dataset, _ = datasets
@@ -39,50 +103,63 @@ def test_multitask_classifier_reload_correctness():
 
     trainer = LightningTorchModel(model=model,
                                   batch_size=16,
+                                  model_dir="test_multitask_restore_dir",
                                   accelerator="cuda",
                                   devices=-1,
                                   log_every_n_steps=1,
-                                  strategy="fsdp",
-                                  fast_dev_run=True)
+                                  strategy="fsdp")
 
     trainer.fit(valid_dataset, nb_epoch=3)
-    # get a some 10 weights for assertion
-    weights = trainer.model.model.layers[0].weight[:10].detach().cpu().numpy()
 
-    trainer.save("multitask_classifier.ckpt")
+    # Store original model state dictionary for comparison
+    original_state_dict = trainer.model.model.state_dict()
 
-    # Reload model and checkpoint
-    reload_model = dc.models.MultitaskClassifier(n_tasks=len(tasks),
-                                                 n_features=1024,
-                                                 layer_sizes=[1000],
-                                                 dropouts=0.2,
-                                                 learning_rate=0.0001,
-                                                 device="cpu",
-                                                 batch_size=16)
+    # Create new model and trainer instance
+    restore_model = dc.models.MultitaskClassifier(n_tasks=len(tasks),
+                                                  n_features=1024,
+                                                  layer_sizes=[1000],
+                                                  dropouts=0.2,
+                                                  learning_rate=0.0001,
+                                                  device="cpu",
+                                                  batch_size=16)
 
-    trainer = LightningTorchModel.reload("multitask_classifier.ckpt",
-                                         model=reload_model,
-                                         batch_size=16,
-                                         accelerator="cuda",
-                                         devices=-1,
-                                         log_every_n_steps=1,
-                                         strategy="fsdp",
-                                         fast_dev_run=True)
+    trainer2 = LightningTorchModel(model=restore_model,
+                                   batch_size=16,
+                                   model_dir="test_multitask_restore_dir",
+                                   accelerator="cuda",
+                                   devices=-1,
+                                   log_every_n_steps=1,
+                                   strategy="fsdp")
 
-    # get a some 10 weights for assertion
-    reloaded_weights = trainer.model.model.layers[0].weight[0][:10].detach(
-    ).cpu().numpy()
+    # Restore from checkpoint
+    trainer2.restore()
 
-    _ = trainer.predict(valid_dataset)
+    # Run prediction to ensure model is properly loaded
+    _ = trainer2.predict(valid_dataset)
 
-    # make it equal with a tolerance of 1e-5
-    assert torch.allclose(torch.tensor(weights),
-                          torch.tensor(reloaded_weights),
-                          atol=1e-5)
+    # Get restored model state dictionary for comparison
+    restored_state_dict = trainer2.model.model.state_dict()
+
+    # Verify that the restored state dict has the same keys
+    assert original_state_dict.keys() == restored_state_dict.keys()
+
+    # Verify that the restored weights are identical to the original weights
+    for key in original_state_dict:
+        torch.testing.assert_close(
+            original_state_dict[key].detach().cpu(),
+            restored_state_dict[key].detach().cpu(),
+            msg=f"Weight mismatch for key {key} after restore operation.")
+
+    # Clean up
+    try:
+        if os.path.exists("test_multitask_restore_dir"):
+            shutil.rmtree("test_multitask_restore_dir")
+    except:
+        pass  # Ignore cleanup errors caused by file locks
 
 
 @pytest.mark.torch
-def test_gcn_model_reload_correctness():
+def test_gcn_model_restore_correctness():
     L.seed_everything(42)
     featurizer = dc.feat.MolGraphConvFeaturizer()
     tasks, all_dataset, _ = dc.molnet.load_bace_classification(featurizer)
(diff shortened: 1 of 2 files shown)


deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,277 @@
...
+            Number of unique classes for classification metrics.
+
+        Returns
+        -------
+        Union[Score, Tuple[Score, Score]]
+            Dictionary mapping metric names to scores. If per_task_metrics is True,
+            returns a tuple of (multitask_scores, all_task_scores).
+        """
+
+        # Process input metrics
```
ARY2260: Why did you reimplement evaluator similar to deepchem [Evaluator class ](https://github.com/deepchem/deepchem/blob/1b2f6b33d2ef5146a0c4000fcdb01958b8c87df7/deepchem/models/models.py#L218) instead of directly using it?

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,277 @@
...
+import logging
+import numpy as np
+import lightning as L
+from rdkit import rdBase
+
+rdBase.DisableLog('rdApp.warning')
+logger = logging.getLogger(__name__)
+
+
+class DeepChemLightningTrainer:
```
riya-singh28: I think you can name this as LightningTorchModel as it's not only responsible for training, but also for inference and eval, and looks like a lightning wrapper for torchmodel

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,277 @@
...
+    >>> predictions = trainer.predict(valid_dataset)
+    >>> trainer.save_checkpoint("model.ckpt")
+    >>> # To reload:
+    >>> trainer2 = DeepChemLightningTrainer.load_checkpoint("model.ckpt", model=model)
+    """
+
+    def __init__(self,
+                 model: TorchModel,
+                 batch_size: int = 32,
+                 **trainer_kwargs: Any) -> None:
```
riya-singh28, line 34: Please document additional lightning **kwargs in the docstring.

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,277 @@
...
+        # Train the model
+        self.trainer.fit(self.lightning_model, data_module, ckpt_path=ckpt_path)
+
+    def predict(self,
+                dataset: Dataset,
+                transformers: List[Transformer] = [],
+                other_output_types: Optional[OneOrMany[str]] = None,
+                num_workers: int = 0,
+                uncertainty: Optional[bool] = None,
+                ckpt_path: Optional[str] = None):
```
riya-singh28, line 217: Can we have a restore method or something similar instead of having a separate ckpt_path argument in fit and predict?

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,277 @@
...
+        **trainer_kwargs
+            Additional trainer arguments.
+
+        Returns
+        -------
+        DeepChemLightningTrainer
+            New trainer instance with loaded model.
+        """
+        # Create trainer first
+        trainer = DeepChemLightningTrainer(model=model,
```
riya-singh28: Why are you using a recursive call of the same class for load_checkpoint? Is that meant to be a replacement for init()?

ARY2260 (changes requested):
I have requested comment on why evaluate is reimplemented?

deepchem/models/lightning/trainer.py
bhuvanmdev: I've added some additional docs to make it more clear.

deepchem/models/lightning/trainer.py
bhuvanmdev: The method that can be considered similar to `restore` is the `load_checkpoint` method.
The ckpt_path arguments used in fit, predict are internally defined in lightning.
So both of these can be used interchangeably for the same funtionality.

deepchem/models/lightning/trainer.py
bhuvanmdev: I've added additional docs to make it more clear. i.e
- Uses `LightningTorchModel's` predict method to get predictions.
- Performs additional concatenation of predictions to ensure correct shape.

bhuvanmdev: Update:- Now it inherits Model base class to use evaluate, as suggested.

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,350 @@
...
+    >>> trainer.save_checkpoint("model.ckpt")
+    >>> # To reload:
+    >>> trainer2 = LightningTorchModel.load_checkpoint("model.ckpt", model=model)
+    """
+
+    def __init__(self,
+                 model: TorchModel,
+                 batch_size: int = 32,
+                 model_dir: Optional[str] = "default_model_dir",
+                 **trainer_kwargs: Any) -> None:
```
ARY2260, line 34: please add docstring for __init__()

ARY2260: you can move it from the main class docstring

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,350 @@
...
+                if not isinstance(cb, ModelCheckpoint)
+            ]
+            callbacks_list.extend(other_callbacks)
+
+        # Create a new trainer with the updated callbacks if we have checkpoint callback
+        if checkpoint_interval > 0:
+            # Update trainer_kwargs to include our callbacks
+            updated_kwargs = self.trainer_kwargs.copy()
+            updated_kwargs['callbacks'] = callbacks_list
+            self.trainer = L.Trainer(**updated_kwargs)
```
ARY2260, line 198: is it ok to reinitialize trainer in fit method? if after init, the fit is called twice, does this not reset trainer state?

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,350 @@
...
+            self.trainer_kwargs['default_root_dir'] = model_dir
+
+        # Create the Lightning module
+        self.lightning_model: DCLightningModule = DCLightningModule(model)
+
+        # Initialize the base Model class with model_dir to ensure compatibility
+        super(LightningTorchModel, self).__init__(model=model,
+                                                  model_dir=model_dir)
+
+    def fit(self,
```
ARY2260, line 122: highlight in the docs that max_epochs in part of main init instead of fit(), which is generally the case in torch model class

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,350 @@
...
+            train_dataset: Dataset,
+            max_checkpoints_to_keep: int = 5,
+            checkpoint_interval: int = 1000,
+            num_workers: int = 4,
+            ckpt_path: Optional[str] = None):
```
ARY2260, line 129: also, in deepchem torch models, fit can be called for 1 specific epoch, hence if there is a restore option within fit to restart fit from latest checkpoint in model_dir. Can't something similar be done here instead? ckpt_path can be given from model_dir if restore = True.

deepchem/models/lightning/trainer.py
ARY2260: similar to my suggestion in fit(), can you directly take latest checkpoint from model_dir()

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,350 @@
...
+        # Train the model
+        self.trainer.fit(self.lightning_model, data_module, ckpt_path=ckpt_path)
+
+    def predict(self,
+                dataset: Dataset,
+                transformers: List[Transformer] = [],
+                other_output_types: Optional[OneOrMany[str]] = None,
+                num_workers: int = 0,
+                uncertainty: Optional[bool] = None,
+                ckpt_path: Optional[str] = None):
```
ARY2260, line 217: does trainer automatically handle if ckpt_path is None?

docs/source/api_reference/models.rst
```diff
@@ -658,12 +658,12 @@ DeepChem supports the use of `PyTorch-Lightning`_ to build PyTorch models.
 
 .. _`PyTorch-Lightning`: https://www.pytorchlightning.ai/
 
-DCLightningModule
```
ARY2260: is DCLightningModule removed from deepchem? If not, then why are you replacing docs of any existing module?

ARY2260 (changes requested):
I have added some suggestions and questions.

deepchem/models/lightning/trainer.py
bhuvanmdev: Yes. Hence it's the default value given.

docs/source/api_reference/models.rst
bhuvanmdev: Ya I think this change wasn't supposed to be in here. Will just add a new one. Thanks for pointing it out.

I have updated it properly now.

deepchem/models/lightning/trainer.py
bhuvanmdev: Yes, the re-init of the trainer itself is not an issue, since the **kwargs are passed as is and also the trainer is a stateless object. The model weights are kept under the model wrapper `DClightningmodel`.

deepchem/models/lightning/trainer.py
bhuvanmdev: Since the trainer is getting initiated at fit, as suggested I moved the epoch parameter to fit() itself along with docs.

deepchem/models/lightning/trainer.py
bhuvanmdev: Yes i have implemented that logic as suggested.

deepchem/models/lightning/trainer.py
bhuvanmdev: done

deepchem/models/lightning/trainer.py
bhuvanmdev: done

deepchem/models/lightning/tests/test_lightning_trainer.py
```diff
@@ -0,0 +1,222 @@
...
+                                  devices=-1,
+                                  log_every_n_steps=1,
+                                  strategy="fsdp",
+                                  fast_dev_run=True)
+
+    trainer.fit(valid_dataset)
+    # get a some 10 weights for assertion
+    weights = trainer.model.model.layers[0].weight[:10].detach().cpu().numpy()
+
+    trainer.save_checkpoint("multitask_classifier.ckpt")
```
rbharath: You should implement a custom save function so the user does not have to do this. Make it something automatic

```
trainer.save(...)
```

We don't have save_checkpoint elsewhere in deepchem. Please make it fit with the standard deepchem API for save/reload

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,381 @@
...
+            Path to a checkpoint file to resume training from. If None, starts fresh.
+
+        Notes
+        -----
+        If `max_checkpoints_to_keep` is set to n, the trainer will keep the last n checkpoints plus the
+        last checkpoint created when the fit ends successfully, named `last.ckpt`.
+        """
+        # Set the number of epochs in the trainer kwargs
+        if 'max_epochs' not in self.trainer_kwargs:
+            self.trainer_kwargs['max_epochs'] = nb_epoch
```
ARY2260: this could become ambiguous if fit is sent with nb_epoch value but its not applied to trainer because max_epochs already in kwargs

avoid epoch based parameter completely in init.

ARY2260: and just set self.trainer_kwargs['max_epochs'] = nb_epoch

ARY2260: or in updated_kwargs

ARY2260 (changes requested):
There is small change request for nb_epoch and rest LGTM

deepchem/models/lightning/trainer.py
bhuvanmdev: Yup. Done.

deepchem/models/lightning/tests/test_lightning_trainer.py
```diff
@@ -0,0 +1,218 @@
...
+                                  devices=-1,
+                                  log_every_n_steps=1,
+                                  strategy="fsdp",
+                                  fast_dev_run=True)
+
+    trainer.fit(valid_dataset, nb_epoch=3)
+    # get a some 10 weights for assertion
+    weights = trainer.model.model.layers[0].weight[:10].detach().cpu().numpy()
+
+    trainer.save("multitask_classifier.ckpt")
```
rbharath: Add some unit tests that use the standard deepchem save/reload mechanism. Please make sure we match the standard deepchem convention

deepchem/models/lightning/tests/test_lightning_trainer.py
riya-singh28, line 1: @bhuvanmdev Add a test case that checks the backward compatibility of trained checkpoints to work with their Torch Model class.
Step 1: Save a GCN checkpoint trained using the save_checkpoint method of Lightning Trainer.
Step 2: Load the checkpoint using the simple restore() method of the TorchModel class.

bhuvanmdev: This does not work out of the box, since the key's names are different in lightning and deepchem checkpoints. I would need to create a separate method, that converts lightning checkpoints into deepchem checkpoints.

bhuvanmdev: Also I would prefer to raise a separate pr for this, since this pr has already become quite big in size 😅.

deepchem/models/lightning/trainer.py
```diff
@@ -0,0 +1,382 @@
...
+
+
+class LightningTorchModel(Model):
+    """A wrapper class that handles the training and inference of DeepChem models using Lightning.
+
+    This class provides a high-level interface for training and running inference
+    on DeepChem models using PyTorch Lightning's training infrastructure. It wraps
+    DeepChem models in Lightning modules and handles data loading, training loops,
+    and checkpoint management. Currently, it supports strategies like DDP (Distributed Data Parallel)
+    and FSDP (Fully Sharded Data Parallel) for distributed training, as well as single-device training.
```
rbharath, line 27: You need to add a detailed discussion of the issue of non-compatibility of checkpoints between deepchem and lightning here.

PR #4587, opened 2025-12-08
ModularTorchModel load_from_pretrained function fix

Harindhar10:
In the load_from_pretrained function of ModularTorchModel, when a checkpoint file is provided as the source of pretrained weights, the target model’s restore function is invoked. As a result, the optimizer state and global step from the source model are also loaded. This appears to be a bug because:

- The restore function was designed for resuming training on the same model, not for transferring weights to a different one.
- Loading the optimizer state (which contains accumulated gradients from the pretraining dataset) and the global step (which shortens early optimizer warmup steps) is inappropriate when fine-tuning on a new dataset and task.
- This issue does not occur when weights are loaded from a model class rather than from a checkpoint file.
 
Code snippet to replicate the bug:
https://colab.research.google.com/drive/1auXGnn6PkwqT8aHTiPMxKJC0uqOm8-hp?usp=sharing

Fix #(issue)

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/torch_models/modular.py
+++ b/deepchem/models/torch_models/modular.py
@@ -325,9 +325,18 @@ class ModularTorchModel(TorchModel):
             self.build_model()
 
         elif source_model is None:
-            self.restore(components=components,
-                         checkpoint=checkpoint,
-                         model_dir=model_dir)
+            logger.info('Loading pretrained model from checkpoint file')
+            if checkpoint is None:
+                checkpoints = sorted(self.get_checkpoints(model_dir))
+                if len(checkpoints) == 0:
+                    raise ValueError('No checkpoint found')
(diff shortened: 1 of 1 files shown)


rbharath:
@riya-singh28 @ARY2260 Please review

riya-singh28:
@Harindhar10 @rbharath 
I went through the code given to replicate the issue, and here is what I think can be an issue. The infograph and infographstar models have different components, and the component 'unsup_encoder' is not present in the infograph. To use load_for_pretrained, the component layer architecture should be present in both the source model and the other model. While you don't see an issue when running the first code, the state_dict is not being updated there.

image

Harindhar10:
Understood. I got the snippet from test_infograph_pretrain_overfit() function

Issue #4912, opened 2026-03-28
Chemberta broken with transformers>=5.0


vivekvar-dl:
Hey, I was setting up DeepChem locally with the latest transformers (5.4.0) and ran into a couple of issues with Chemberta:

**1. Import crash**

```python
from deepchem.models.torch_models.chemberta import Chemberta
```

Fails with:
```
ModuleNotFoundError: No module named 'transformers.models.roberta.tokenization_roberta_fast'
```

This is because `transformers.models.roberta.tokenization_roberta_fast` was renamed to `tokenization_roberta_old` in transformers 5.x. The fix is straightforward — switching to `from transformers import RobertaTokenizerFast` works across both 4.x and 5.x. I have a PR for this.

**2. Tokenizer loading failure (deeper issue)**

Even after fixing the import, `RobertaTokenizerFast.from_pretrained()` fails when loading old-format tokenizers (vocab.json + merges.txt):

```
ValueError: not enough values to unpack (expected 2, got 1)
```

This affects both the default Hub tokenizer (`seyonec/PubChem10M_SMILES_BPE_60k`) and locally trained tokenizers saved with `ByteLevelBPETokenizer.save_model()`. The new format (`tokenizer.json` saved via `.save()`) works fine.

This seems like a broader transformers 5.x compat issue that might affect MoLFormer and the HuggingFaceModel base class examples too.

**Environment:**
- deepchem 2.5.0 (installed from source, master branch)
- transformers 5.4.0
- tokenizers 0.22.2
- Python 3.12.3

PR #4913, opened 2026-03-28
fix: update Chemberta imports for transformers 5.x compat

vivekvar-dl:
Fixes the import-time crash from #4912.

`transformers.models.roberta.tokenization_roberta_fast` was removed in transformers 5.x (renamed to `tokenization_roberta_old`), so `from deepchem.models.torch_models.chemberta import Chemberta` throws a `ModuleNotFoundError`.

This switches to the top-level `from transformers import RobertaTokenizerFast` which works across 4.x and 5.x. Also updated the modeling imports for consistency.

There's a separate runtime issue with `RobertaTokenizerFast` in 5.x where old-format tokenizer files don't load — tracked in #4912.

## Type of change

## Checklist

- [x] New and existing imports work with transformers 4.x and 5.x
- [x] No breaking changes to existing functionality

### DIFF
--- a/deepchem/models/torch_models/chemberta.py
+++ b/deepchem/models/torch_models/chemberta.py
@@ -1,9 +1,8 @@
 from typing import Dict, Any, Tuple
 from deepchem.models.torch_models.hf_models import HuggingFaceModel
-from transformers.models.roberta.modeling_roberta import (
-    RobertaConfig, RobertaForMaskedLM, RobertaForSequenceClassification)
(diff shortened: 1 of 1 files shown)


gotnochill815-web:
Tested this with transformers 5.x — import fix works 
However, ChemBERTa still fails during initialization due to tokenizer incompatibility with legacy BPE formats.
Opening a follow-up PR to ensure full runtime compatibility.

PR #4992, opened 2026-06-11
Added generate() function in hf_models.py and added 2 test files (Cleaned Version)

priyadp1:
## Description

I added a generate() function in hf_models.py and added a test file called test_hf.py. I did this to support text generation via Hugging Face models. My implementation wraps the HuggingFace generate() API and includes device handling. Essentially, you can run my code on a CPU or GPU. My test_hf.py was added to verify that generated outputs are returned as decoded strings.
I ran the full DeepChem test suite (pytest) and verified that all existing tests pass with this implementation.

## Type of change

- [X ] New feature (non-breaking change which adds functionality)

## Checklist

- [ X] My code follows [the style guidelines of this project](https://deepchem.readthedocs.io/en/latest/development_guide/coding.html)
  - [X ] Run `yapf -i <modified file>` and check no errors (**yapf version must be  0.32.0**)
  - [ X] Run `flake8 <modified file> --count` and check no errors
  - [X ] Run `python -m doctest <modified file>` and check no errors
- [X ] I have performed a self-review of my own code
- [X ] I have commented my code, particularly in hard-to-understand areas
- [X ] I have added tests that prove my fix is effective or that my feature works
- [X ] New unit tests pass locally with my changes
- [ X] I have checked my code and corrected any misspellings

### DIFF
--- a/deepchem/models/torch_models/hf_models.py
+++ b/deepchem/models/torch_models/hf_models.py
@@ -8,10 +8,11 @@ import numpy as np
 import torch
 from deepchem.models.optimizers import LearningRateSchedule
 from deepchem.models.torch_models import TorchModel
+from deepchem.data import Dataset
 from deepchem.trans import Transformer, undo_transforms
 from deepchem.utils.typing import LossFn, OneOrMany
 from transformers.data.data_collator import DataCollatorForLanguageModeling
-from transformers.models.auto import AutoModel, AutoModelForSequenceClassification, AutoModelForMaskedLM, AutoModelForUniversalSegmentation
+from transformers.models.auto import AutoModel, AutoModelForSequenceClassification, AutoModelForMaskedLM, AutoModelForUniversalSegmentation, AutoModelForCausalLM
 
 logger = logging.getLogger(__name__)
 
@@ -238,6 +239,10 @@ class HuggingFaceModel(TorchModel):
             elif self.task == "universal_segmentation":
                 self.model = AutoModelForUniversalSegmentation.from_pretrained(
                     model_dir, trust_remote_code=True, **self.config)
+            elif self.task == 'causal_lm':
+                self.model = AutoModelForCausalLM.from_pretrained(
+                    model_dir, trust_remote_code=True,
+                    **self.config).to(self.device)
             else:
                 self.model = AutoModel.from_pretrained(model_dir,
                                                        trust_remote_code=True,
@@ -277,6 +282,10 @@ class HuggingFaceModel(TorchModel):
 
     def _prepare_batch(self, batch: Tuple[Any, Any, Any]):
         smiles_batch, y, w = batch
+        if self.tokenizer.pad_token is None:
+            self.tokenizer.pad_token = self.tokenizer.eos_token
+            if hasattr(self.model, "config"):
+                self.model.config.pad_token_id = self.model.config.eos_token_id
         tokens = self.tokenizer(smiles_batch[0].tolist(),
                                 padding=True,
                                 return_tensors="pt")
@@ -288,6 +297,15 @@ class HuggingFaceModel(TorchModel):
                 'input_ids': inputs.to(self.device),
                 'labels': labels.to(self.device),
                 'attention_mask': tokens['attention_mask'].to(self.device),
+            }
+            return inputs, None, w
+        elif self.task == 'causal_lm':
+            input_ids = tokens['input_ids'].to(self.device)
+            attention_mask = tokens['attention_mask'].to(self.device)
+            inputs = {
+                'input_ids': input_ids,
+                'attention_mask': attention_mask,
+                'labels': input_ids.clone()
             }
             return inputs, None, w
         elif self.task in ['regression', 'classification', 'mtr']:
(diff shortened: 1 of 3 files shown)


riya-singh28:
@priyadp1 Please fix the failing flake8 test. To check locally - Run flake8 <modified file> --count and check no errors
Screenshot from 2026-06-17 13-18-28
Screenshot from 2026-06-17 13-18-17

deepchem/models/torch_models/hf_models.py
```diff
@@ -636,3 +654,71 @@ def fill_mask(self,
...
+        """Generate text using HuggingFace's text generation pipeline.
+
+        Parameters
+        ----------
+        dataset : Dataset
+            A dataset in DeepChem format whose X field contains input sequences to
+            condition generation on.
+        **kwargs:
+            Additional keyword arguments to pass to HuggingFace's generate
+            method.
```
riya-singh28: please include a list of common kwargs for huggingface generate and add a link to its docs as well.

deepchem/models/torch_models/hf_models.py
```diff
@@ -636,3 +654,71 @@ def fill_mask(self,
...
+        if self.tokenizer.pad_token is None:
+            self.tokenizer.pad_token = self.tokenizer.eos_token
+            self.model.config.pad_token_id = self.model.config.eos_token_id
+
+        if hasattr(self.tokenizer, "padding_side"):
+            self.tokenizer.padding_side = "left"
+        device = self.device
+
+        original_device = next(self.model.parameters()).device
+        if device.type == 'mps':
```
riya-singh28, line 702: If this works fine for the new MPS users, as we discussed in the last meeting, can you add a condition check and update the device only for users with Mac version issues?

riya-singh28: @shreyasvinaya Can you please check if this code works fine for your Mac device?

deepchem/models/torch_models/hf_models.py
```diff
@@ -238,6 +239,10 @@ def load_from_pretrained(  # type: ignore
             elif self.task == "universal_segmentation":
                 self.model = AutoModelForUniversalSegmentation.from_pretrained(
                     model_dir, trust_remote_code=True, **self.config)
+            elif self.task == 'causal_lm':
+                self.model = AutoModelForCausalLM.from_pretrained(
+                    model_dir, trust_remote_code=True,
+                    **self.config).to(self.device)
```
riya-singh28: Why do you need to set self.model device here itself? It's handled differently for other task types, right?

deepchem/models/torch_models/hf_models.py
```diff
@@ -277,6 +282,10 @@ def load_from_pretrained(  # type: ignore
 
     def _prepare_batch(self, batch: Tuple[Any, Any, Any]):
         smiles_batch, y, w = batch
+        if self.tokenizer.pad_token is None:
+            self.tokenizer.pad_token = self.tokenizer.eos_token
```
riya-singh28: Is this change necessary here? Is it correct to set tokenizer pad_token to tokenizer eos_token? I am concerned because _prepare_batch is used by several HuggingFace-derived models.

Can you please add some relevant references as well in the thread here?

riya-singh28 (changes requested):
@priyadp1 I have requested a few changes.

deepchem/models/torch_models/hf_models.py
priyadp1: @riya-singh28 Done

deepchem/models/torch_models/hf_models.py
priyadp1: @riya-singh28 fixed it to have mac os version <14, then move to CPU, else stay on mps

deepchem/models/torch_models/hf_models.py
priyadp1: @riya-singh28 It was a typo, removed it

deepchem/models/torch_models/hf_models.py
priyadp1: @riya-singh28 For causal-lm tasks, especially for decoder models like GPT2, they lack a padding token. So, the padding is needed for decoder-only models in general, even though models like LLaMA have a padding token. I took that line out from prepare_ batch and moved it to the init, since that's where the task setup is. I then fixed up the causal_lm part of the prepare batch by masking the padding positions to -100. This is important because then the labels are the same as the inputs with the padding tokens ignored.

Reference 1: https://huggingface.co/docs/transformers/main_classes/data_collator (Masking padding positions to -100)

Reference 2: https://github.com/huggingface/transformers/issues/8452  (GPT2 padding)

deepchem/models/torch_models/hf_models.py
priyadp1: @riya-singh28  Link: https://huggingface.co/docs/transformers/main_classes/text_generation

deepchem/models/torch_models/hf_models.py
sudo-rsingh: @priyadp1 Is there a more elegant way of doing this which we dont have to hardcode a version number, that would be much better because if someday this package increases the requirements to > 18 or something then the test will get broken.

riya-singh28 (approved):
@rbharath LGTM.

rbharath (approved):
LGTM

PR #4994, opened 2026-06-12
Add RFDiffusion input embedding layers

chidwipak:
## Description

Adds the three input embedding layers that the RFDiffusion denoiser uses to
convert raw backbone coordinates and timesteps into feature vectors:

- `SinusoidalTimestepEmbedding` – turns an integer diffusion timestep into a
  sinusoidal vector so the model knows how noisy the current input is.
- `ResidueEmbedding` – projects the 9 backbone coordinates (N, CA, C × xyz)
  into a higher-dimensional feature space before the transformer.
- `PositionalEncoding` – adds sinusoidal chain-position info so the model can
  tell residues apart by their order in the sequence.

This is PR A in a 4-part split. None of these classes import from later PRs so
they stand on their own. PR B (CosineSchedule) and PR C (BackboneDiffusion)
will build on top of this file.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - In this case, we recommend to discuss your modification on GitHub issues before creating the PR
- [ ] Documentations (modification for documents)

## Checklist

- [x] My code follows [the style guidelines of this project](https://deepchem.readthedocs.io/en/latest/development_guide/coding.html)
  - [ ] Run `yapf -i <modified file>` and check no errors (**yapf version must be  0.32.0**)
  - [ ] Run `mypy -p deepchem` and check no errors
  - [x] Run `flake8 <modified file> --count` and check no errors
  - [ ] Run `python -m doctest <modified file>` and check no errors
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New unit tests pass locally with my changes
- [x] I have checked my code and corrected any misspellings

PR #4995, opened 2026-06-12
Add RFDiffusion cosine noise schedule

chidwipak:
## Description

Adds `CosineSchedule`, the diffusion noise schedule used by the RFDiffusion
protein backbone model.

The schedule implements the cosine variance schedule from Improved DDPM, which
tends to produce better samples than a linear schedule because the noise is
added more gradually at the start and end of the trajectory.

Three methods are exposed:

- `q_sample` – adds noise to clean data at a given timestep (the "forward" pass used during training).
- `p_sample` – denoises by one step using a model's noise prediction (the "reverse" pass used at inference).
- `sample` – runs the full reverse loop from noise to clean data.

All schedule tensors are stored as buffers so they move with the model when
you call `.to(device)` and are saved/loaded with checkpoints.

**Depends on PR #4994** (the embedding layers). This is PR B in a 4-part split.
PR C (BackboneDiffusion) and PR D (RFDiffusionModel) will build on top.

## Type of change

## Checklist

### DIFF
--- a/deepchem/models/torch_models/rfdiffusion.py
+++ b/deepchem/models/torch_models/rfdiffusion.py
@@ -10,15 +10,18 @@ denoiser transformer sees them:
   each with x, y, z) into a higher-dimensional space.
 - ``PositionalEncoding`` — adds sinusoidal position information so the
   network knows where each residue sits in the chain.
-
-None of these have learnable parameters that depend on later modules, so
-they are shipped separately as the first small, reviewable PR.
+- ``CosineSchedule`` — implements the improved cosine variance schedule
+  for the forward (noise-adding) and reverse (denoising) diffusion steps.
 
 References
 ----------
 .. [1] Watson, J. L., et al. "De novo design of protein structure and function
    with RFdiffusion." Nature 620.7976 (2023): 1089-1100.
-.. [2] Vaswani, A., et al. "Attention is all you need." NeurIPS 2017.
+.. [2] Ho, J., Jain, A., & Abbeel, P. "Denoising diffusion probabilistic
+   models." NeurIPS 2020.
+.. [3] Nichol, A. Q., & Dhariwal, P. "Improved denoising diffusion
+   probabilistic models." ICML 2021.
+.. [4] Vaswani, A., et al. "Attention is all you need." NeurIPS 2017.
 
 Notes
 -----
@@ -26,10 +29,12 @@ This module requires PyTorch to be installed.
 """
 
 import math
+from typing import Optional, Tuple
 
 try:
     import torch
     import torch.nn as nn
+    import torch.nn.functional as F
 except ModuleNotFoundError:
     raise ImportError('These classes require PyTorch to be installed.')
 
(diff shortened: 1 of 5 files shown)


deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    >>> schedule = CosineSchedule(num_timesteps=100)
+    >>> import torch
+    >>> x0 = torch.randn(2, 10, 9)
+    >>> t = torch.tensor([0, 50])
+    >>> noisy_x, noise = schedule.q_sample(x0, t)
+    >>> noisy_x.shape
+    torch.Size([2, 10, 9])
+    """
+
+    def __init__(self, num_timesteps: int = 1000, s: float = 0.008) -> None:
```
JoseAntonioSiguenza, line 241: Is there an specific reason for setting default to 0.008? Does authors use this as default?

JoseAntonioSiguenza: I checked "Improved DDPM" paper, and `s` comes from pixel bin size of 1/127.5 ~ 0.00784 (by dividing the normalized range from -1 to 1 (range of 2) by 255 intervals). Please add this hyperparameter value to the documentation.

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        super().__init__()
+        self.num_timesteps = num_timesteps
```
JoseAntonioSiguenza, line 243: Please add `num_timesteps` refers to "T" term from eq. 17 in reference paper

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        alpha_cumprod = torch.cos((t + s) / (1 + s) * math.pi / 2)**2
+        alpha_cumprod = alpha_cumprod / alpha_cumprod[0]
+        betas = 1 - (alpha_cumprod[1:] / alpha_cumprod[:-1])
+        betas = torch.clamp(betas, 0, 0.999)
+
+        alphas = 1.0 - betas
+        alpha_cumprod = torch.cumprod(alphas, dim=0)
+        alpha_cumprod_prev = F.pad(alpha_cumprod[:-1], (1, 0), value=1.0)
+
+        self.register_buffer('betas', betas)
```
JoseAntonioSiguenza: I'm not sure about using `self.register_buffer` (I dont remember using it or see ir anywhere else in deepchem). I dont think this matches Deepchem's API. Is there an specific reason for registering buffer?

JoseAntonioSiguenza: I'd suggest creating a field to access from `q_sample`, `p_sample`, etc methods

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        Parameters
+        ----------
+        x0 : torch.Tensor
+            Clean data of shape ``(batch, ...)``.
+        t : torch.Tensor
+            Integer timesteps of shape ``(batch,)``.
+        noise : torch.Tensor, optional
+            Pre-sampled noise tensor. Drawn from N(0, I) if not provided.
+
+        Returns
```
JoseAntonioSiguenza, line 293: Please add ussage examples

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        sqrt_one_minus = self.sqrt_one_minus_alpha_cumprod[t_idx].to(x0.device)
+
+        while sqrt_alpha.dim() < x0.dim():
+            sqrt_alpha = sqrt_alpha.unsqueeze(-1)
+            sqrt_one_minus = sqrt_one_minus.unsqueeze(-1)
+
+        noisy_x = sqrt_alpha * x0 + sqrt_one_minus * noise
+        return noisy_x, noise
+
+    @torch.no_grad()
```
JoseAntonioSiguenza: I dont think `@torch.no_grad()` decorator matches Deepchem API. Instead please make it explicit inside method

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+        Parameters
+        ----------
+        model : nn.Module
+            The denoiser network. Called as ``model([x_t, t])``.
+        x_t : torch.Tensor
+            Noisy data of shape ``(batch, num_residues, coord_dim)``.
+        t : torch.Tensor
+            Current timestep of shape ``(batch,)``.
+
```
JoseAntonioSiguenza, line 336: Please add usage examples

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        Returns
+        -------
+        torch.Tensor
+            Denoised data at timestep t-1, same shape as x_t.
+        """
+        device = x_t.device
+        noise_pred = model([x_t, t])
+
+        t_idx = t.to(self.betas.device)
```
JoseAntonioSiguenza: Can't we use `device` from ln 327? Since, `q_sample` also uses `.to(device)` lets add `self.device` in __init__ method.

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        while nonzero_mask.dim() < x_t.dim():
+            nonzero_mask = nonzero_mask.unsqueeze(-1)
+
+        var = self.posterior_variance[t_idx].to(device)
+        while var.dim() < x_t.dim():
+            var = var.unsqueeze(-1)
+
+        return posterior_mean + nonzero_mask * torch.sqrt(var) * noise
+
+    @torch.no_grad()
```
JoseAntonioSiguenza: Same as above

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        ----------
+        model : nn.Module
+            The denoiser network.
+        shape : tuple of int
+            Shape of samples to produce: ``(batch, num_residues, coord_dim)``.
+        device : torch.device
+            Device to generate on.
+
+        Returns
+        -------
```
JoseAntonioSiguenza, line 401: Please add usage examples

deepchem/models/torch_models/rfdiffusion.py
```diff
@@ -185,3 +190,194 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        var = self.posterior_variance[t_idx].to(device)
+        while var.dim() < x_t.dim():
+            var = var.unsqueeze(-1)
+
+        return posterior_mean + nonzero_mask * torch.sqrt(var) * noise
+
+    @torch.no_grad()
+    def sample(self, model: nn.Module, shape: Tuple[int, ...],
+               device: torch.device) -> torch.Tensor:
+        """Generate samples by running the full reverse diffusion loop.
```
JoseAntonioSiguenza, line 389: Please add how this will be used in RFDiffusion model

deepchem/models/torch_models/tests/test_rfdiffusion_schedule.py
```diff
@@ -0,0 +1,86 @@
+"""Tests for the RFDiffusion cosine noise schedule."""
```
JoseAntonioSiguenza, line 1: Please add brief docstrings to unit tests describing the purpose of each test (what errors ahould it catch, what are the expected outputs, etc)

deepchem/models/torch_models/tests/test_rfdiffusion_schedule.py
```diff
@@ -0,0 +1,86 @@
...
+    has_torch = False
+
+
+@pytest.mark.torch
+@pytest.mark.skipif(not has_torch, reason="PyTorch not installed")
+class TestCosineSchedule:
+
+    def test_q_sample_shape(self):
+        schedule = CosineSchedule(num_timesteps=100)
+        x0 = torch.randn(2, 10, 9)
```
JoseAntonioSiguenza, line 21: Lets add a test from real image data instead of random data. Can we test similar behaviour as original implementation? https://github.com/openai/improved-diffusion/tree/main/datasets

JoseAntonioSiguenza:
Nice implementation! I left some comments related to Deepchem API and tests.

deepchem/models/torch_models/tests/test_rfdiffusion_schedule.py
JoseAntonioSiguenza: Lets test proper behaviour against real image data. I think if we succesfully this test, the implementation would be more robust and safe.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: > Is there an specific reason for setting default to 0.008? Does authors use this as default?

Added that context to the docstring. `0.008` is the default used in the Improved DDPM paper.

chidwipak: > I checked "Improved DDPM" paper, and `s` comes from pixel bin size of 1/127.5 ~ 0.00784 (by dividing the normalized range from -1 to 1 (range of 2) by 255 intervals). Please add this hyperparameter value to the documentation.

> Added that note next to `s`. The docstring now calls out the `1 / 127.5 ≈ 0.00784` bin-size motivation for the `0.008` default.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: Added that to the parameter docs. `num_timesteps` now points to the `T` term from equation 17.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: > I'm not sure about using `self.register_buffer` (I dont remember using it or see ir anywhere else in deepchem). I dont think this matches Deepchem's API. Is there an specific reason for registering buffer?

Kept `register_buffer` here because these tensors are fixed schedule state, not learnable weights. They still need to move with the module and be written to the checkpoint.

chidwipak: > I'd suggest creating a field to access from `q_sample`, `p_sample`, etc methods

These tensors are already exposed as fields through `register_buffer`, so the methods read them as `self.betas`, `self.alpha_cumprod`, and the rest.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: Added a short `q_sample()` example to the docstring.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: Changed both methods to use `with torch.no_grad():` inside the body so they follow the usual DeepChem style.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: Added a small `p_sample()` example with a dummy denoiser so the call pattern is clear.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: I kept the device on the tensors instead of adding a separate `self.device` field. These buffers already move with the module, and `q_sample()` only casts the indexed values to match the input tensor.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: I kept the same pattern in `sample()`. The schedule buffers already track the module device, so there is not a second device field to keep in sync.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: Added a short `sample()` example to the docstring.

deepchem/models/torch_models/rfdiffusion.py
chidwipak: Added a short note on how `RFDiffusionModel` uses `q_sample()` during training and `sample()` during generation.

deepchem/models/torch_models/tests/test_rfdiffusion_schedule.py
chidwipak: Added brief docstrings to the schedule tests so each one says what behavior it is checking.

deepchem/models/torch_models/tests/test_rfdiffusion_schedule.py
chidwipak: I kept the test inputs synthetic and added a reference check against the Improved DDPM cosine formula instead. Since the schedule only depends on the timestep grid, that is closer to the original implementation than swapping in image tensors.

deepchem/models/torch_models/rfdiffusion.py
JoseAntonioSiguenza: Yes, but my suggestion is to add the needed tensors(self.betas, self.alpha_cumprod, etc) explictily as fields instead of using `self.register_buffer`

JoseAntonioSiguenza:
LGTM

rbharath (approved):
LGTM

PR #5009, opened 2026-06-18
Fix dropped samples in SingletaskStratifiedSplitter

### LINKED ISSUE #5008: Bug: Remaining samples are silently discarded in `SingletaskStratifiedSplitter.split()
### Description

The implementation of `SingletaskStratifiedSplitter.split()` appears to silently discard the remaining samples that do not fit into a complete block during stratified splitting.

In `splitters.py` (around line 778), the code handles the remaining indices as follows:

```python
# Append remaining examples to train
if sortidx.shape[0] > 0:
    np.hstack([train_idx, sortidx])
```

However, the result of `np.hstack()` is not assigned back to `train_idx`.

As a consequence, the remaining indices are never included in any of the returned train, validation, or test splits.

### Expected Behavior

All samples from the original dataset should be assigned to exactly one split.

The following invariant should always hold:

```python
len(train) + len(valid) + len(test) == len(dataset)
```

### Actual Behavior

Up to 9 remaining samples can be silently omitted from all splits.

### Impact

This causes data loss during dataset splitting and can lead to:

* Reduced training data size
* Inconsistent experiment results
* Violations of user expectations regarding dataset integrity
* Downstream pipeline inconsistencies when split sizes do not sum to the original dataset size

### Steps to Reproduce

```python
import numpy as np
import deepchem as dc

X = np.random.rand(95, 10)
y = np.random.rand(95, 1)

splitter = dc.splits.SingletaskStratifiedSplitter()

train, valid, test = splitter.split(
    dataset,
    frac_train=0.8,
    frac_valid=0.1,
    frac_test=0.1
)

print("Original dataset size:", len(dataset))
print(
    "Split sizes - Train:",
    len(train),
    "Valid:",
    len(valid),
    "Test:",
    len(test)
)

print(
    "Lost samples:",
    len(dataset) - (len(train) + len(valid) + len(test))
)
```

Example output:

```text
Original dataset size: 95
Split sizes - Train: 72 Valid: 9 Test: 9
Lost samples: 5
```

## Root Cause

The remaining indices are passed to `np.hstack()`, but the returned array is discarded:

```python
np.hstack([train_idx, sortidx])
```

Since NumPy operations do not modify arrays in place, the remaining indices are never added to `train_idx` and are omitted from all returned splits.

## Suggested Fix

Assign the result of `np.hstack()` back to `train_idx`:

```python
train_idx = np.hstack([train_idx, sortidx])
```

This ensures that all remaining samples are included in the training split and that every sample in the original dataset is assigned to exactly one output split.

Debajeet-1411:
Fixes #5008
## Descripion

This PR fixes a bug in `SingletaskStratifiedSplitter.split()` where the remaining samples that do not fit into a complete block are silently discarded.

The issue occurs because the result of `np.hstack([train_idx, sortidx])` is computed but never assigned back to `train_idx`. As a result, the remaining indices are omitted from all returned splits.

## Changes Made

In `splitters.py`, the following code:

```python
if sortidx.shape[0] > 0:
    np.hstack([train_idx, sortidx])
```

was changed to:

```python
if sortidx.shape[0] > 0:
    train_idx = np.hstack([train_idx, sortidx])
```

## Why This Change Is Needed

The original implementation computed the concatenation of the remaining samples with the training indices but discarded the returned array because it was not assigned back to any variable.

Since NumPy array operations are not performed in place, the remaining samples were never included in any split.

## Impact

This fix ensures that:

* No samples are silently dropped during splitting.
* Every sample from the original dataset is assigned to exactly one output split.
* The total number of samples across train, validation, and test splits matches the original dataset size.

## Verification

Using a dataset whose size is not divisible by 10:

```python
X = np.random.rand(95, 10)
y = np.random.rand(95, 1)

splitter = dc.splits.SingletaskStratifiedSplitter()
```

Before this fix:

```text
Original dataset size: 95
Split sizes - Train: 72 Valid: 9 Test: 9
Lost samples: 5
```

After this fix:

```text
Original dataset size: 95
Split sizes - Train: 77 Valid: 9 Test: 9
Lost samples: 0
```
### Verification Video

https://github.com/user-attachments/assets/a2c5a7fb-79bd-4b42-9dd9-7eb168a75f99

### DIFF
--- a/deepchem/splits/splitters.py
+++ b/deepchem/splits/splitters.py
@@ -775,7 +775,7 @@ class SingletaskStratifiedSplitter(Splitter):
 
         # Append remaining examples to train
         if sortidx.shape[0] > 0:
-            np.hstack([train_idx, sortidx])
+            train_idx = np.hstack([train_idx, sortidx])
 
         return (train_idx, valid_idx, test_idx)
 


siddhant-shahhh (commented):
Nice catch. The diagnosis is right, `np.hstack` isn't in place, so the remainder was computed and thrown away.

Ran your fix across a few dataset sizes, coverage problem is gone:

    n=9     returned=9     lost=0
    n=15    returned=15    lost=0
    n=19    returned=19    lost=0
    n=37    returned=37    lost=0
    n=101   returned=101   lost=0
    n=137   returned=137   lost=0

Two things I hit while checking, in case you want to fold them in.

**1. The index arrays are float64, so they still can't be used on a NumpyDataset.**

Your repro calls `split()` and stops there, which works. One step further, passing the indices to `select()`, which is what `train_valid_test_split` and `train_test_split` do internally still fails with your fix applied:

    tr, va, te = splitter.split(dataset, 0.8, 0.1, 0.1, seed=0)
    tr.dtype                 # float64
    dataset.select(tr)       # IndexError: arrays used as indices must be of
                             # integer (or boolean) type

`train_idx`, `valid_idx` and `test_idx` start as `np.array([])`, which is float64, and hstacking integer indices onto it keeps it float64. `NumpyDataset.select` indexes `self.X[indices]` directly so it rejects them; `DiskDataset.select` is a separate shard-based path and tolerates them, which is probably why this hasn't come up.

`dtype=int` fixes it in the same block you're already touching:

    train_idx: np.ndarray = np.array([], dtype=int)
    valid_idx: np.ndarray = np.array([], dtype=int)
    test_idx: np.ndarray = np.array([], dtype=int)

**2. Worth adding a test and there's a reason this went unnoticed.**

`grep -rn "SingletaskStratified" --include="*.py" .` turns up no test that exercises this class. `test_singletask_stratified_split` in `deepchem/splits/tests/test_splitter.py` is named for it but instantiates `dc.splits.ScaffoldSplitter()`. Its final assertion is

    assert sorted(merged_dataset.ids) == sorted(solubility_dataset.ids)

which is exactly the invariant that would have caught this, just pointed at the wrong class. Point it at `SingletaskStratifiedSplitter` with a dataset size that isn't a multiple of 10 and your fix is locked in. n=10 and n=100 pass either way, which is also why the docstring example never showed it.

Tested on deepchem 2.8.1.dev, master @455d07f3e, Python 3.12.

PR #5044, opened 2026-07-06
Fix seed parameter in ScaffoldSplitter and resolve type issues

### LINKED ISSUE #5043: Bug: `seed` Parameter Ignored in `ScaffoldSplitter.split()`
## Description

The `seed` parameter passed to the `split()` method of `ScaffoldSplitter` is completely ignored. As a result, when splitting datasets where multiple scaffold groups have the same number of molecules (which is very common), ties are always resolved in the same order.

This prevents users from obtaining different deterministic randomized splits when changing the seed.

---

### Steps

1. Create a `ScaffoldSplitter` instance.
2. Call `.split()` or `.train_test_split()` on a dataset using different random seeds (for example, `seed=42` and `seed=100`).
3. Observe that the resulting train, validation, and test indices are identical regardless of the seed.

### Minimal Reproducible Example

```python
import deepchem as dc
import numpy as np

# Create a dummy dataset with duplicate/same-size scaffold groups
smiles = [
    "CC(C)Cl",
    "CCC(C)CO",
    "CCCCCCCO",
    "CCCCCCCC(=O)OC"
]

dataset = dc.data.DiskDataset.from_numpy(
    X=np.zeros(len(smiles)),
    y=np.ones(len(smiles)),
    w=np.zeros(len(smiles)),
    ids=smiles
)

splitter = dc.splits.ScaffoldSplitter()

# Split using different seeds
train_1, valid_1, test_1 = splitter.split(dataset, seed=42)
train_2, valid_2, test_2 = splitter.split(dataset, seed=100)

# The splits are identical despite different seeds
assert train_1 == train_2
assert valid_1 == valid_2
assert test_1 == test_2
```

---

## Expected Behavior

The splitter should use the provided `seed` to deterministically randomize scaffold groups that have the same size.

Specifically:

- The same seed should always produce the same split.
- Different seeds should produce different split assignments whenever there are scaffold groups with identical sizes.
- Overall scaffold-based splitting behavior should remain deterministic and reproducible.

---

## Actual Behavior

The output is identical for every seed value because the `seed` argument is never used during the splitting process.

---

## Root Cause

Inside `deepchem/splits/splitters.py`, the `ScaffoldSplitter.split()` method accepts a `seed` argument:

```python
def split(
    self,
    dataset: Dataset,
    frac_train: float = 0.8,
    frac_valid: float = 0.1,
    frac_test: float = 0.1,
    seed: Optional[int] = None,
    log_every_n: Optional[int] = 1000
) -> Tuple[List[int], List[int], List[int]]:
```

However, the `seed` parameter is never passed to any random number generator or used to shuffle scaffold groups before assigning them to the train, validation, and test sets.

Instead, scaffold groups are sorted only by size (and existing ordering), causing ties to always be resolved identically.

---

## Suggested Fix

Before assigning scaffold groups to dataset splits:

1. Group scaffold sets by their size.
2. Initialize a deterministic random number generator using the provided `seed`.
3. Shuffle scaffold groups within each equal-sized bucket.
4. Continue the existing greedy assignment algorithm.

Example:

```python
rng = np.random.default_rng(seed)

# Group scaffold sets by length
groups = defaultdict(list)
for scaffold_set in scaffold_sets:
    groups[len(scaffold_set)].append(scaffold_set)

# Shuffle only equal-sized scaffold groups
ordered_sets = []
for size in sorted(groups.keys(), reverse=True):
    rng.shuffle(groups[size])
    ordered_sets.extend(groups[size])
```

This preserves the current scaffold splitting strategy while making tie-breaking deterministic and dependent on the provided seed.

---

## Environment

- **Operating System:** Windows 10
- **Python:** 3.11.7
- **DeepChem:** 2.8.1.dev
- **RDKit:** 2026.03.3

---

## Impact

- The documented `seed` argument has no effect.
- Users cannot generate alternative reproducible scaffold splits by changing the seed.
- Cross-validation and repeated experiments using scaffold splitting become less flexible because every run produces the same tie resolution.

---

## Summary

The `ScaffoldSplitter.split()` API exposes a `seed` parameter, but it is currently unused. Introducing deterministic shuffling of equal-sized scaffold groups using the provided seed would make the parameter functional without altering the existing scaffold-based splitting algorithm.

Kanika0306:
## Summary

This PR fixes an issue where the `seed` parameter passed to `ScaffoldSplitter.split()` was accepted but never used.

Previously, scaffold groups were sorted only by size. When multiple scaffold groups had the same number of molecules, ties were always resolved in the same order, producing identical splits regardless of the provided seed.

This change introduces deterministic tie-breaking by shuffling scaffold groups of equal size using a random number generator initialized with the user-provided seed. As a result:

- The same seed always produces the same scaffold split.
- Different seeds can produce different, yet reproducible, scaffold splits when equal-sized scaffold groups are present.
- Existing scaffold-based splitting logic remains unchanged apart from tie resolution.

Fixes #5043

---

### `deepchem/splits/splitters.py`

- Group scaffold sets by their size using `collections.defaultdict`.
- Use `np.random.RandomState(seed)` to deterministically shuffle scaffold groups within each size bucket.
- Preserve the existing greedy scaffold assignment algorithm after tie-breaking.

---

## Type of Change

- [x] Bug fix (non-breaking change)

---

## Testing

The change was verified by running the existing splitter test suite.

- ✅ `pytest splits/tests/test_scaffold_splitter.py`
- ✅ `pytest splits/tests/test_splitter.py`
- ✅ `mypy splits/splitters.py`

All tests passed successfully with no type checking issues.

**Test Results**

Screenshot 2026-07-06 232542
Screenshot 2026-07-06 232631

## Notes

This change only affects the ordering of scaffold groups with identical sizes during splitting. It does not modify the overall scaffold splitting strategy or change behavior for scaffold groups with unique sizes.

### DIFF
--- a/deepchem/splits/splitters.py
+++ b/deepchem/splits/splitters.py
@@ -1552,7 +1552,9 @@ class ScaffoldSplitter(Splitter):
             Each indices is a list of integers.
         """
         np.testing.assert_almost_equal(frac_train + frac_valid + frac_test, 1.)
-        scaffold_sets = self.generate_scaffolds(dataset)
+        scaffold_sets = self.generate_scaffolds(
+            dataset,
+            log_every_n=log_every_n if log_every_n is not None else 1000)
 
         from collections import defaultdict
         size_to_group_idxs = defaultdict(list)
--- a/deepchem/splits/splitters.py
+++ b/deepchem/splits/splitters.py
@@ -1555,14 +1555,16 @@ class ScaffoldSplitter(Splitter):
         scaffold_sets = self.generate_scaffolds(dataset)
 
         from collections import defaultdict
-        rng = np.random.RandomState(seed)
         size_to_group_idxs = defaultdict(list)
         for i, group in enumerate(scaffold_sets):
             size_to_group_idxs[len(group)].append(i)
-        for idxs in size_to_group_idxs.values():
-            rng.shuffle(idxs)
 
         ordered_sizes = sorted(size_to_group_idxs.keys(), reverse=True)
+        if seed is not None:
+            rng = np.random.RandomState(seed)
+            for size in ordered_sizes:
+                rng.shuffle(size_to_group_idxs[size])
+
         scaffold_sets = [
             scaffold_sets[i]
             for size in ordered_sizes
--- a/deepchem/splits/splitters.py
+++ b/deepchem/splits/splitters.py
@@ -1442,8 +1442,8 @@ def _split_fingerprints(fps: List, size1: int,
     remaining_fp = fps[1:]
     remaining_indices = list(range(1, len(fps)))
     max_similarity_to_group = [
-        DataStructs.BulkTanimotoSimilarity(fps[0], remaining_fp),
-        [0] * len(remaining_fp)
+        np.array(DataStructs.BulkTanimotoSimilarity(fps[0], remaining_fp)),
+        np.zeros(len(remaining_fp))
     ]
     # Return identity if no tuple to split to
     if size2 == 0:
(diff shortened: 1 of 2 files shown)


Kanika0306:
@bharatr21 @Zrahay could you kindly review and guide me

PR #5068, opened 2026-07-23
Warn when create_dataset is called on a non-empty directory

### LINKED ISSUE #3402: Warning or error before saving a dataset to a directory with pre-existing files

ArneshBanerjee:
Fixes #3402

`DiskDataset.create_dataset` now logs a warning when the target `data_dir`
already exists and has files in it. Before this, saving to a non-empty
directory would silently overwrite everything there.

The overwrite still proceeds so existing behaviour is unchanged. Added
`test_create_dataset_warns_on_existing_files` in `test_datasets.py` to
verify the warning is logged.

### DIFF
--- a/deepchem/data/datasets.py
+++ b/deepchem/data/datasets.py
@@ -1266,6 +1266,13 @@ class DiskDataset(Dataset):
             data_dir = tempfile.mkdtemp()
         elif not os.path.exists(data_dir):
             os.makedirs(data_dir)
+        else:
+            existing_files = os.listdir(data_dir)
+            if existing_files:
+                logger.warning(
+                    "The directory %s already contains %d file(s): %s. "
+                    "Saving to this directory will overwrite existing data.",
+                    data_dir, len(existing_files), existing_files)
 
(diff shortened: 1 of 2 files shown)


ArneshBanerjee:
Rebased on master.

Small usability fix: `DiskDataset.create_dataset` silently overwrites whatever is already in `data_dir`, so pointing it at a populated directory loses data with no indication. This logs a warning in that case and leaves behaviour otherwise unchanged.

CI doesn't appear to have run on this yet. Could a maintainer trigger it and take a look?

PR #5073, opened 2026-07-25
Add RFDiffusion SO(3) rotation diffusion utilities

chidwipak:
This adds the rotational part of the RFDiffusion noise process as a new
self-contained module, `rfdiffusion_so3.py`. It only depends on torch.

What it adds:

- `so3_exp_map` / `so3_log_map` - Rodrigues exponential and a
  quaternion-based logarithm. Routing the log map through a quaternion keeps
  it stable for rotations near `pi`, where reading the angle straight from
  the trace loses precision.
- `log_beta_schedule` - the log-linear sigma schedule used for rotational
  diffusion.
- `IGSO3` - the isotropic Gaussian on SO(3). It caches the PDF, CDF, and
  score over an omega grid and samples rotations by inverse CDF. The
  truncated heat-kernel series (2000 terms), the omega grid
  (`linspace(0, pi, N+1)[1:]`), and the autograd score all match upstream
  RFdiffusion's defaults.
- `so3_reverse_step` - one reverse-time Euler-Maruyama step on SO(3).

Testing:

- Exp/log roundtrip across angles, including close to `pi`.
- Schedule shape, endpoints, and monotonicity.
- IGSO(3) normalisation, CDF monotonicity, and the score checked against a
  finite difference of `log f(omega)`.
- Sampling shapes, valid rotation outputs, and angle statistics.
- Reverse-step shape, determinism, and input validation.

Doctests and the test module both pass locally, and the module is
flake8/yapf clean.

Part of the RFDiffusion series. The module is standalone, so it can be
reviewed on its own.

### DIFF
--- a/deepchem/models/torch_models/rfdiffusion_so3.py
+++ b/deepchem/models/torch_models/rfdiffusion_so3.py
@@ -1,0 +1,610 @@
+"""SO(3) diffusion utilities for RFDiffusion.
+
+This module implements the rotational component of the RFDiffusion noise
+process [Watson2023]_: an IGSO(3) (Isotropic Gaussian on SO(3)) diffusion
+with a logarithmic variance schedule and a reverse-time Euler-Maruyama
+integrator on the so(3) tangent space [Leach2022]_, [Yim2023]_.
+
+Conventions
+-----------
+* Rotations are represented as ``(..., 3, 3)`` proper rotation matrices
+  R in SO(3) with det(R) = +1.
+* A rotation is parameterised on the tangent space so(3) =~ R^3 by a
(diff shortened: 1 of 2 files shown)


Shriprasad-P (commented):
## Review

**PR:** Add RFDiffusion SO(3) rotation diffusion utilities

**Touched:** `deepchem/models/torch_models/rfdiffusion_so3.py`, `deepchem/models/torch_models/tests/test_rfdiffusion_so3.py`

- Scoped to the stated goal; please confirm CI is green.
- Size: +971/-0.

Drive-by review after reading the diff.

PR #5076, opened 2026-07-28
Add RFDiffusion contig parser and motif scaffolding utilities

chidwipak:
This adds the contig parsing and motif-scaffolding utilities that RFDiffusion's
inpainting pipeline uses, in a new self-contained module
`deepchem/utils/rfdiffusion_contigs.py` (only depends on numpy).

What it adds:

- `parse_contig_string` - parses the RFDiffusion CLI contig dialect
  (e.g. `5-10/A12-30/5-10`) into motif and linker segments.
- `ContigMap` / `RealisedContig` - the parsed layout and a concrete
  realisation with sampled linker lengths, plus the motif mask and the
  per-position mapping back to reference residues.
- `build_motif_mask` - aligns a realised layout with reference coordinates and
  returns the coordinate array and a boolean motif mask.
- `freeze_motif_coords` - keeps motif atoms fixed at the reference geometry
  during sampling.

The contig grammar matches the dialect accepted by the official RFDiffusion CLI
(motif segments like `A12-30`, variable linkers like `10-20`, and fixed-length
linkers like `8`, separated by `/` or whitespace).

Tests cover segment validation, string parsing (motif, linker, fixed-length,
whitespace/slash separators, and error cases), length ranges and realisation,
the motif mask / source mapping, coordinate alignment against a reference, and
the freeze helper. Doctests and the test module both pass locally, and the
module is flake8/yapf clean.

This is the last core RFDiffusion PR. The module is standalone, so it can be
reviewed on its own.

### DIFF
--- a/deepchem/utils/rfdiffusion_contigs.py
+++ b/deepchem/utils/rfdiffusion_contigs.py
@@ -47,16 +47,6 @@ try:
 except ModuleNotFoundError:
     raise ImportError('rfdiffusion_contigs requires NumPy to be installed.')
 
-__all__ = [
-    'LinkerSegment',
-    'MotifSegment',
-    'ContigMap',
-    'RealisedContig',
-    'parse_contig_string',
-    'build_motif_mask',
-    'freeze_motif_coords',
-]
-
 _MOTIF_RE = re.compile(r'^([A-Za-z])(-?\d+)-(-?\d+)$')
 _LINKER_RE = re.compile(r'^(\d+)-(\d+)$')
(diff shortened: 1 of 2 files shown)


Shriprasad-P (commented):
## Review

**PR:** Add RFDiffusion contig parser and motif scaffolding utilities

**Touched:** `deepchem/utils/rfdiffusion_contigs.py`, `deepchem/utils/test/test_rfdiffusion_contigs.py`

- Looks like a concrete bugfix based on the title/diff.
- Size: +742/-0.

Drive-by review after reading the diff.

PR #5094, opened 2026-08-12
Fix AtomicConv applying dropout during eval

### LINKED ISSUE #5092: AtomicConv applies dropout during eval, so predict() is nondeterministic
## 🐛 Bug

`AtomicConv.forward` calls `F.dropout(x, dropout)` without passing `training`. `F.dropout` defaults to `training=True`, so it stays active after `model.eval()`. `dropouts` defaults to `0.5`, and `AtomConvModel.predict()` runs through this path, so repeated predictions on the same input differ.

`deepchem/models/torch_models/layers.py`:

```python
if dropout > 0:
    x = F.dropout(x, dropout)
```

Other modules guard this. `fcnet.py` uses `if dropout > 0.0 and self.training`, and the uncertainty path uses an explicit `dropout_switch`. `AtomicConv` has neither.

## To Reproduce

1. Build an `AtomConvModel` with a nonzero `dropouts` value (0.5 is the default).
2. Fit it, then call `predict` on the same dataset twice.
3. The two prediction arrays differ.

Isolating the mechanism:

```python
import torch, torch.nn as nn, torch.nn.functional as F
m = nn.Linear(8, 8).eval()
t = torch.ones(1, 8)
print(torch.equal(F.dropout(m(t), 0.5), F.dropout(m(t), 0.5)))
# False
print(torch.equal(F.dropout(m(t), 0.5, training=m.training),
                  F.dropout(m(t), 0.5, training=m.training)))
# True
```

## Expected behavior

Dropout is disabled under `eval()`, so `predict()` returns the same values for the same input.

## Environment

* OS: macOS 26.5.1
* Python version: 3.12.11
* DeepChem version: main @ 43e6f2e
* PyTorch version: 2.8.0
* Any other relevant information: same defect class as #5086 / PR #5087, which fixed `ScScore.forward`

## Additional context

Both this and the layer chaining question in #5091 sit in the same loop. Filing separately since they are independent.

chiruu12:
`AtomicConv.forward` calls `F.dropout(x, dropout)` without passing `training`. `F.dropout` defaults to `training=True`, so dropout stays active after `model.eval()`. `dropouts` defaults to `0.5` and `AtomConvModel.predict()` runs through this path, so repeated predictions on the same input differ.

Other modules already guard this. `fcnet.py` uses `if dropout > 0.0 and self.training`, and the uncertainty path uses an explicit `dropout_switch`. `AtomicConv` had neither.

Passes `training=self.training` so `eval()` disables dropout, and adds a test that runs two forward passes under `eval()` on the same input and asserts the outputs are identical.

Closes #5092.

### DIFF
--- a/deepchem/models/torch_models/layers.py
+++ b/deepchem/models/torch_models/layers.py
@@ -2801,7 +2801,7 @@ class AtomicConv(nn.Module):
             x = layer(inputs_x)
 
             if dropout > 0:
-                x = F.dropout(x, dropout)
+                x = F.dropout(x, dropout, training=self.training)
 
             if activation_fn is not None:
                 x = activation_fn(x)
--- a/deepchem/models/torch_models/tests/test_acnn.py
+++ b/deepchem/models/torch_models/tests/test_acnn.py
@@ -175,8 +175,7 @@ def test_atomic_convolution_module_eval_dropout():
                              size=(batch_size, num_atoms, max_num_neighbors))
         nbrs_z = torch.randint(1,
                                10,
-                               size=(batch_size, num_atoms,
-                                     max_num_neighbors))
+                               size=(batch_size, num_atoms, max_num_neighbors))
         return coords, nbrs, nbrs_z
 
     frag1 = _frag(frag1_num_atoms)
--- a/deepchem/models/torch_models/tests/test_acnn.py
+++ b/deepchem/models/torch_models/tests/test_acnn.py
@@ -147,3 +147,51 @@ def test_atomic_convolution_model_variable():
     preds = atomic_convnet.predict(train)
     assert preds.shape == (1, 1)
     assert np.count_nonzero(preds) > 0
+
+
+@pytest.mark.torch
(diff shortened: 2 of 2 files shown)


chiruu12:
Anything you need from me on this one?

`AtomicConv.forward` calls `F.dropout(x, dropout)` without `training`, which defaults to
True, so `eval()` does not turn dropout off and two predictions on the same input differ.
`fcnet.py` already guards this the normal way. Happy to rebase.

Shriprasad-P (commented):
## Review

**PR:** Fix AtomicConv applying dropout during eval

**Touched:** `deepchem/models/torch_models/layers.py`, `deepchem/models/torch_models/tests/test_acnn.py`

- Addresses a concrete correctness/reliability issue based on the title and diff.
- Includes or touches tests — good for locking the behavior.
- Size looks manageable (+48/-1).

Commenting as a drive-by reviewer after reading the diff. Happy to look again if maintainers want a deeper pass on a specific file.

PR #5100, opened 2026-08-19
Fix log(0) numerical instability in PoissonLoss and CategoricalCrossEntropy

### LINKED ISSUE #5099: PoissonLoss and CategoricalCrossEntropy – log(0) causes -inf / NaN in PyTorch path
## 🐛 Bug

`PoissonLoss` and `CategoricalCrossEntropy` call `torch.log()` directly on raw model outputs **without any numerical guard**. When the output tensor contains `0.0` (e.g., a fully saturated softmax class or a zero predicted count), `torch.log(0)` evaluates to `-inf`, which propagates through the loss → gradient becomes `NaN` → model weights silently corrupt.

This is inconsistent with sibling loss classes in the same file (`ShannonEntropy`, `VAE_KLDivergence`) that already apply an epsilon guard.

**Affected classes in `deepchem/models/losses.py`:**

| Class | Line | Unsafe Expression |
|---|---|---|
| `PoissonLoss._create_pytorch_loss` | 158 | `torch.log(output)` |
| `CategoricalCrossEntropy._create_pytorch_loss` | 206 | `torch.log(output)` |

## To Reproduce

1. Install DeepChem with PyTorch.
2. Run the following snippet:

```python
import torch
from deepchem.models.losses import PoissonLoss, CategoricalCrossEntropy

# Case 1: PoissonLoss with a zero predicted count
loss_fn = PoissonLoss()._create_pytorch_loss()
output = torch.tensor([0.0, 1.0, 2.0])  # 0.0 is a valid Poisson prediction
labels = torch.tensor([1.0, 1.0, 1.0])
result = loss_fn(output, labels)
print("PoissonLoss result:", result)
# → tensor(nan) or tensor(-inf)

# Case 2: CategoricalCrossEntropy with a saturated zero-probability class
loss_fn2 = CategoricalCrossEntropy()._create_pytorch_loss()
output2 = torch.tensor([[0.0, 0.5, 0.5]])  # first class probability is 0
labels2 = torch.tensor([[1.0, 0.0, 0.0]])  # but it is the true class
result2 = loss_fn2(output2, labels2)
print("CategoricalCrossEntropy result:", result2)
# → tensor([nan]) or tensor([inf])
```

3. Observe that both results are non-finite (`nan` or `inf`). Any downstream `.backward()` call will corrupt all model parameters with no error raised.

## Expected behavior

Both loss functions should return a **finite scalar/tensor** for any valid input, matching the behavior of `ShannonEntropy` and `VAE_KLDivergence` which already handle this correctly in the same file:

```python
# ShannonEntropy (line 466) — correct pattern already in codebase
torch.log(1e-20 + inputs)

# VAE_KLDivergence (line 414) — correct pattern already in codebase
torch.log(1e-20 + torch.square(logvar))
```

### Proposed fix

Clamp the argument to `torch.log` to a small positive floor using `torch.finfo`, which is dtype-aware (`float16` ≈ `6e-8`, `float32` ≈ `1.2e-7`, `float64` ≈ `2.2e-16`):

```python
# PoissonLoss._create_pytorch_loss  (line 158)
eps = torch.finfo(output.dtype).eps
return torch.mean(output - labels * torch.log(output.clamp(min=eps)))

# CategoricalCrossEntropy._create_pytorch_loss  (line 206)
eps = torch.finfo(output.dtype).eps
return -torch.sum(labels * torch.log(output.clamp(min=eps)), dim=-1)
```

A regression test should also be added to `deepchem/models/tests/test_losses.py`:

```python
import torch
from deepchem.models.losses import PoissonLoss, CategoricalCrossEntropy

def test_poisson_loss_zero_output_no_nan():
    loss_fn = PoissonLoss()._create_pytorch_loss()
    output = torch.tensor([0.0, 1.0, 2.0])
    labels = torch.tensor([1.0, 1.0, 1.0])
    result = loss_fn(output, labels)
    assert torch.isfinite(result), f"PoissonLoss returned non-finite value: {result}"

def test_categorical_crossentropy_zero_prob_no_nan():
    loss_fn = CategoricalCrossEntropy()._create_pytorch_loss()
    output = torch.tensor([[0.0, 0.5, 0.5]])
    labels = torch.tensor([[1.0, 0.0, 0.0]])
    result = loss_fn(output, labels)
    assert torch.all(torch.isfinite(result)), f"CategoricalCrossEntropy returned non-finite value: {result}"
```

## Environment

* OS:
* Python version:
* DeepChem version:
* RDKit version (optional):
* TensorFlow version (optional):
* PyTorch version:
* Any other relevant information:

## Additional context

- The bug is **completely silent** — no exception is raised. The first symptom is `loss=nan` appearing in training logs after one unlucky batch, which is often misdiagnosed as a learning rate or architecture issue.
- The **TensorFlow paths are not affected** — `_compute_tf_loss` for both classes delegates to `tf.keras.losses`, which applies its own internal clamping.
- The fix is a 2-line change per class, making this a great first-contribution PR.

mishradwaterlaw:
## Description

Fix #5099 

**Summary of changes:**
This PR fixes a critical numerical instability bug in `PoissonLoss` and `CategoricalCrossEntropy` (PyTorch implementations). Previously, both classes called `torch.log()` directly on the raw model output. If the model predicted exactly `0.0` (e.g., a zero-probability class from a saturated softmax, or a zero count prediction), `torch.log(0)` would evaluate to `-inf`, which propagated through the loss and silently corrupted the model's weights with `NaN` gradients. 

**Changes made:**
- Clamped the `output` tensor to a dtype-aware minimum epsilon floor (`torch.finfo(output.dtype).eps`) before calling `torch.log()` in `PoissonLoss` and `CategoricalCrossEntropy`. 
- This approach matches the numerical stability safeguards already in place for sibling classes like `ShannonEntropy` and `VAE_KLDivergence` in the same module.
- Added comprehensive regression tests (`TestNumericalStabilityLosses`) in `test_losses.py` covering edge cases with `0.0` probabilities/predictions for both loss functions and verifying they are finite across different dtypes.

## Type of change

## Checklist


siddhant-shahhh (commented):
The instability is real, but I don't think finfo(...).eps is the right constant, and clamp has
a side effect that works against what the PR is trying to do.

eps is machine epsilon (1.19e-07 for float32), not the smallest positive float
(tiny = 1.18e-38). So the clamp fires on ordinary probabilities that float32 represents exactly:

    p=1e-04   exact CE  9.21034   patched  9.21034
    p=1e-06   exact CE 13.81551   patched 13.81551
    p=1e-08   exact CE 18.42068   patched 15.94238
    p=1e-10   exact CE 23.02585   patched 15.94238

That's 7 nats off at p=1e-10 on an input that was never degenerate.

The bigger issue is the gradient. clamp is flat below its min, so the model gets no signal
precisely where it is most wrong:

    p        dL/dp exact    clamp(min=eps)   log(x+eps)
    1e-4     -1.0e+04       -1.0e+04         -9988
    1e-9     -1.0e+09        0               -8.32e+06

Same for PoissonLoss. At output=0 the patched gradient is +0.333, so the optimizer pushes the
prediction further down, deeper into the clamped region. The exact gradient 1 - l/o goes to
-inf, i.e. pushes it up. log(output + eps) gives -2.8e+06 there: finite, and the right sign.

Small correction to the docstrings: they say log(0) produces NaN gradients.
output - labels*log(output) at output=0 is +inf when labels > 0. It's NaN only when
labels == 0, from 0 * -inf.

So I'd suggest torch.log(output + eps) instead of clamping, and finfo(dtype).tiny if you want
the floor as low as it goes.

Separately, `yapf -d` flags losses.py (line 167 is too long for the google style CI uses), so
the formatting job will fail as-is.

PR #5120, opened 2026-09-04
fix: pass training=self.training to F.dropout in AtomicConv

### LINKED ISSUE #5092: AtomicConv applies dropout during eval, so predict() is nondeterministic

KaiOnCode:
## Description

Fixes #5092: AtomicConv applies dropout during eval, so predict() is nondeterministic

The `F.dropout()` call in `AtomicConv.forward()` was missing the `training` parameter, causing dropout to remain active during `model.eval()` mode. This made predictions nondeterministic for the same input.

## Changes

Changed `F.dropout(x, dropout)` to `F.dropout(x, dropout, training=self.training)` in the `AtomicConv.forward()` method to respect the model's training mode.

## Testing

The fix follows the same pattern used in other modules (e.g., `fcnet.py`) which guard dropout with `if dropout > 0.0 and self.training`.

## Related

- Similar to #5086 / PR #5087 which fixed the same issue in `ScScore.forward()`


KaiOnCode:

PR #5121, opened 2026-09-04
fix: pass training=self.training to F.dropout in ScScore

### LINKED ISSUE #5086: ScScore.forward() applies dropout during eval() — F.dropout() missing training=self.training

KaiOnCode:
## Description

Fixes #5086: ScScore.forward() applies dropout during eval() — F.dropout() missing training=self.training

The `F.dropout()` calls in `ScScore.forward()` were missing the `training` parameter, causing dropout to remain active during `model.eval()` mode. This made predictions nondeterministic for the same input.

## Changes

Changed `F.dropout(x, p=self.dropout)` to `F.dropout(x, p=self.dropout, training=self.training)` in both dropout call sites in `ScScore.forward()`.

## Testing

The fix follows the same pattern used in other modules which guard dropout with the training flag.

## Related

- Same issue class as #5092 in `AtomicConv`


KaiOnCode:

PR #5122, opened 2026-09-04
fix: use proper format string in ButinaSplitter logger

### LINKED ISSUE #5082: Malformed logger call in ButinaSplitter.split raises string-formatting error
## Description

`ButinaSplitter.split` logs the clustering cutoff with a malformed `logger` call:

https://github.com/deepchem/deepchem/blob/master/deepchem/splits/splitters.py#L1238

```python
logger.info("Performing butina clustering with cutoff of", self.cutoff)
```

The format string has no `%` placeholder, yet a second positional argument (`self.cutoff`) is passed. The `logging` module defers interpolation and treats trailing positional arguments as `%`-style format args. When the record is emitted it raises `TypeError: not all arguments converted during string formatting`, which the logging framework routes to `handleError` and prints as a traceback on stderr. The cutoff value the message was meant to show is never logged.

## Steps to reproduce

Run any Butina split with logging at INFO level:

```python
import logging
logging.basicConfig(level=logging.INFO)
import numpy as np
import deepchem as dc

smiles = ['C', 'CC', 'CCC', 'CCCC', 'CCCCC']
ds = dc.data.DiskDataset.from_numpy(X=np.zeros(len(smiles)), ids=smiles)
dc.splits.ButinaSplitter().train_test_split(ds)
```

A `--- Logging error ---` traceback with `TypeError: not all arguments converted during string formatting` is printed to stderr.

## Expected

The cutoff value is interpolated and logged without error.

KaiOnCode:
## Description

Fixes #5082: Malformed logger call in ButinaSplitter.split raises string-formatting error

The `logger.info()` call was missing the `%` format placeholder, causing `TypeError` when the logging framework tried to interpolate the cutoff value.

## Changes

Changed `logger.info("Performing butina clustering with cutoff of", self.cutoff)` to `logger.info("Performing butina clustering with cutoff of %s", self.cutoff)`

## Testing

The fix uses the standard `%s` format placeholder for the logging module, which is the recommended approach for lazy string formatting in Python logging.


KaiOnCode:

PR #5123, opened 2026-09-04
fix: correct HingeLoss docstring label values

### LINKED ISSUE #5017: HingeLoss docstring specifies incorrect label values
### HingeLoss docstring specifies incorrect label values

**File:** `deepchem/models/losses.py`

The `HingeLoss` docstring currently states:

> The 'output' argument should contain logits, and all elements of 'labels' should equal 0 or 1.

However, this does not match the implementation.

The TensorFlow implementation uses:

```python
tf.keras.losses.hinge(labels, output)
```

and the PyTorch implementation computes:

```python
torch.clamp(1 - labels * output, min=0)
```

Both correspond to the standard hinge loss formulation, which expects labels in `{-1, 1}`.

As a result, the current documentation is misleading and suggests label values that are inconsistent with the implemented loss function.

### Steps to Reproduce

```python
import torch
from deepchem.models.losses import HingeLoss

loss_fn = HingeLoss()._create_pytorch_loss()

output = torch.tensor([[5.0]], requires_grad=True)
label = torch.tensor([[0.0]])

loss = loss_fn(output, label)
print(loss)
```

The loss is computed using:

```python
1 - labels * output
```

Which is consistent with the standard hinge loss formulation based on labels in `{-1, 1}` rather than `{0, 1}`.
The documentation specifies labels should be 0 or 1, while the implementations use the standard hinge loss formulation that expects labels in {-1, 1}.
### Expected Behavior

The docstring should be updated to state that labels must be `-1` or `1`.

### Screenshot

Image

KaiOnCode:
## Description

Fixes #5017: HingeLoss docstring specifies incorrect label values

The docstring stated labels should equal 0 or 1, but the standard hinge loss formulation (and the implementation) expects labels in {-1, 1}.

## Changes

Updated the docstring to correctly state that labels should equal -1 or 1.

## Background

The PyTorch implementation uses `torch.clamp(1 - labels * output, min=0)` which is the standard hinge loss formulation. The TensorFlow implementation uses `tf.keras.losses.hinge(labels, output)` which also expects labels in {-1, 1}.


KaiOnCode:

PR #5125, opened 2026-09-04
fix: add epsilon guard to torch.log in PoissonLoss and CategoricalCrossEntropy

### LINKED ISSUE #5099: PoissonLoss and CategoricalCrossEntropy – log(0) causes -inf / NaN in PyTorch path

KaiOnCode:
## Description

Fixes #5099: PoissonLoss and CategoricalCrossEntropy – log(0) causes -inf / NaN in PyTorch path

Added epsilon guard to `torch.log` calls to prevent `-inf`/`NaN` when output contains zeros.

## Changes

- `PoissonLoss._create_pytorch_loss`: Added `eps = torch.finfo(output.dtype).eps` and changed `torch.log(output)` to `torch.log(output.clamp(min=eps))`
- `CategoricalCrossEntropy._create_pytorch_loss`: Added `eps = torch.finfo(output.dtype).eps` and changed `torch.log(output)` to `torch.log(output.clamp(min=eps))`

## Rationale

This matches the pattern already used in `ShannonEntropy` and `VAE_KLDivergence` in the same file. The TensorFlow paths are not affected as they delegate to `tf.keras.losses` which applies its own internal clamping.


KaiOnCode:

PR #5128, opened 2026-09-08
Fix DAGLayer/DAGGather rejecting glorot_normal/xavier_normal init

### LINKED ISSUE #5127: DAGLayer and DAGGather reject glorot_normal/xavier_normal init
## 🐛 Bug

DAGLayer and DAGGather both document `init="glorot_normal"` and
`init="xavier_normal"` as supported options, but passing either raises
ValueError instead of constructing the layer.

## To Reproduce

1. `from deepchem.models.torch_models.layers import DAGLayer`
2. `DAGLayer(init="glorot_normal")`
3. Raises `ValueError: Unsupported init: glorot_normal`

Same happens with `DAGGather(init="glorot_normal")`, and with
`init="xavier_normal"` on both classes.

## Expected behavior

Layer should construct normally, using `nn.init.xavier_normal_` for the
weights, same as it already does for `glorot_uniform`/`xavier_uniform`.

## Environment

* OS: Linux
* Python version: 3.12
* DeepChem version: 2.8.1.dev (current master)
* RDKit version (optional): N/A, not needed to reproduce
* TensorFlow version (optional): N/A, not needed to reproduce
* PyTorch version (optional): 2.x
* Any other relevant information: bug is in `_initialize_weights` for both
  classes in deepchem/models/torch_models/layers.py

## Additional context

Looks like a comparison bug — `self.init` (a str) is compared with `==`
against a list instead of checking membership with `in`, so that branch
can never match. Happy to submit a fix if useful.

sumitjhadev:
Fix #5127

DAGLayer and DAGGather were raising `ValueError: Unsupported init` for
`glorot_normal` and `xavier_normal`, even though both are documented as
supported initialization options. The bug was in `_initialize_weights`, which
compared `self.init` (a string) to a list with `==` instead of checking
membership with `in`. Therefore, the normal-initializer branch could never be
reached. This change fixes both classes and adds regression tests for the two
previously broken initializers.

## Type of change

## Checklist

- [x] I have added tests that prove my fix is effective
- [x] Flake8 passes for the modified files
- [x] The focused DAGLayer/DAGGather doctests pass
- [x] No unrelated files were modified


bobby404-K:
Tested this locally on Windows (Python 3.10, PyTorch 2.x).

- Confirmed master was raising `ValueError: Unsupported init: glorot_normal` on layer instantiation due to the string-list equality check.
- Verified that with the membership check update (`in`), both `DAGLayer` and `DAGGather` instantiate properly with `glorot_normal` and `xavier_normal`.
- Verified tensor weights initialize as finite without NaNs.

**Minor suggestion:** Consider adding a test case asserting that an invalid string (e.g., `init='unsupported'`) still raises `pytest.raises(ValueError)` to keep the error-handling path regression-tested.

sumitjhadev:
Good catch @bobby404-K  added test_DAG_layer_invalid_init and test_DAG_gather_invalid_init
in the latest commit (4de207a) to keep that path covered. Thanks for testing
this independently. 

Would appreciate another look when you get a chance.

Shriprasad-P (commented):
Nice focused PR. From a first pass the logic seems sound; happy to see a green CI run and a short note on how this was validated.

PR #5051, opened 2026-07-07
fix(tests): skip pysam and tensorflow tests if missing and resolve scipy.misc.face deprecation

Kanika0306:
## Summary

This PR fixes several issues in the auto-generated `rdkit-stubs` package that prevented static analysis tools from correctly parsing RDKit type stubs.

The generated stubs contained syntax errors, indentation issues, invalid function signatures, and missing symbol re-exports. Since DeepChem extensively imports `rdkit.Chem` and `rdkit.Chem.AllChem`, these issues caused failures in type checking and degraded IDE support.

**Fixes #5049**

---

## Type of Change

- [x] Bug fix (non-breaking change)

---

### Fixed syntax and indentation issues
- Corrected indentation errors in `rdchem.pyi`.
- Fixed invalid function signatures that violated Python's parameter ordering rules.
- Replaced malformed generated type annotations with valid Python typing constructs where appropriate.

### Improved generated type signatures
Updated several generated stubs that incorrectly represented C++ template types.

Files updated include:

- `Chem/rdchem.pyi`
- `Chem/rdMolDescriptors.pyi`
- `Chem/rdmolfiles.pyi`
- `Chem/Scaffolds/rdScaffoldNetwork.pyi`
- `Chem/rdfiltercatalog/__init__.pyi`

### Restored missing exports
- Added the missing re-export of `GetMorganFingerprintAsBitVect` in `Chem/AllChem.pyi` so static type checkers recognize the API correctly.

---

## Why this change is needed

These issues caused several developer-facing problems:

- Static type checkers (`mypy`, `pyright`) failed while parsing RDKit stubs.
- CI pipelines relying on type checking could fail before analyzing project code.
- IDE features such as autocomplete, hover information, and diagnostics were degraded.
- Missing exports generated false-positive type errors even though the corresponding runtime APIs existed.

This PR resolves those issues without changing any runtime behavior.

---

## Testing

The updated stubs were validated by running static analysis and verifying that the affected modules can now be parsed successfully.

---

## Notes

These changes only affect the generated type stubs used by static analysis tools. They do **not** modify RDKit runtime functionality or DeepChem's execution behavior.

---

## Review

cc: @bharatr21 @Zrahay

Could you kindly review these changes and share any feedback or suggestions for improvement? Thank you!

PR #5131, opened 2026-09-12
Fix genomics tests erroring instead of skipping when TensorFlow/simdna are missing

### LINKED ISSUE #5130: Tests in test_genomics.py error instead of skipping when TensorFlow or simdna are missing
## 🐛 Bug

Three tests in deepchem/metrics/tests/test_genomics.py fail rather than skip when
TensorFlow or simdna are not installed.

There are two separate causes.

1. test_in_silico_mutagenesis_shape and test_in_silico_mutagenesis_nonzero both call
   create_model_for_mutagenesis(), which uses tf.keras at line 54:

       NameError: name 'tf' is not defined

   The file imports tensorflow in a try/except at the top and sets has_tensorflow,
   but has_tensorflow is never used anywhere in the file. Both tests carry
   @pytest.mark.tensorflow, which selects them in CI but does not skip them when
   TensorFlow is absent.

2. test_get_motif_scores calls get_motif_scores(), which raises:

       ImportError: This function requires simdna to be installed.

   There is no simdna guard in the test file at all.

test_get_pssm_scores passes, as it only uses numpy and scipy.

## To Reproduce

1. Install DeepChem from source on master (455d07f3e) with `pip install -e .`
2. Do not install tensorflow or simdna
3. Run `python -m pytest deepchem/metrics/tests/test_genomics.py`

Result:

    FAILED test_get_motif_scores
    FAILED test_in_silico_mutagenesis_nonzero
    FAILED test_in_silico_mutagenesis_shape
    3 failed, 1 passed

## Expected behavior

Tests that need an optional dependency should be skipped when it is absent, not
error. This is the same class of problem being fixed for the loader tests in #5051,
in a file that PR does not cover.

## Environment

* OS: macOS 26.6.2 (arm64)
* Python version: 3.12.13
* DeepChem version: 2.8.1.dev (editable install from master @ 455d07f3e)
* RDKit version (optional): 2026.3.6
* TensorFlow version (optional): not installed
* PyTorch version (optional): 2.14.0
* Any other relevant information: simdna not installed

## Additional context

Related but not duplicates: #4745 change the bare `except:` in
this file to `except Exception:` as part of repo-wide sweeps, but none adds a skip
guard, so the tests still error. #3078 lists simdna among the optional dependencies
to trim guarding this test is a small step in that direction.

Happy to open a PR adding @unittest.skipIf guards for both dependencies.

sumitjhadev:
Fix #5130

`test_get_motif_scores`, `test_in_silico_mutagenesis_shape`, and
`test_in_silico_mutagenesis_nonzero` in `deepchem/metrics/tests/test_genomics.py`
were erroring instead of skipping when TensorFlow or simdna were not installed.

There were two separate root causes, both in the test file itself (not in
`genomic_metrics.py`, which already guards its own simdna import correctly):

1. `has_tensorflow` was computed at import time but never checked anywhere.
   The two mutagenesis tests were only decorated with `@pytest.mark.tensorflow`,
   which is a pytest marker for test *selection* (`-m tensorflow` /
   `-m "not tensorflow"`) — it does not skip a test at runtime when the
   dependency is absent. Without TensorFlow installed, `create_model_for_mutagenesis()`
   raised `NameError: name 'tf' is not defined`.
2. There was no guard at all for `simdna`. `test_get_motif_scores` called
   `get_motif_scores()` unconditionally, which raises
   `ImportError: This function requires simdna to be installed.` when simdna
   is absent, and nothing in the test caught that to skip.

This PR adds a `has_simdna` detection block (mirroring the existing
`has_tensorflow` pattern, but using the narrower `except ModuleNotFoundError`
already used by `genomic_metrics.py` itself), and adds
`@unittest.skipIf` guards to the three affected tests — additive to the
existing `@pytest.mark.tensorflow` markers, which are left in place.

No changes to `genomic_metrics.py` or any other file. No dependencies added.

### Verification

Without TensorFlow or simdna installed:
- Before: `3 failed, 1 passed`
- After: `1 passed, 3 skipped`

With simdna installed, TensorFlow still absent:
- `test_get_motif_scores` now runs and **passes** (not just skipped),
  confirming the guard un-gates correctly rather than blanket-skipping:
  `2 passed, 2 skipped`

`pytest -m tensorflow` with TensorFlow absent now skips cleanly instead of
erroring: `2 skipped, 2 deselected`.

Full `deepchem/metrics/tests/` suite: `26 passed, 2 skipped`, no regressions.

## Type of change

## Checklist

  - [x] Run `yapf -i <modified file>` and check no errors (ran with yapf 0.43.0 locally — no diff produced; happy to re-run against the pinned 0.32.0 if CI flags anything)
  - [x] Run `mypy -p deepchem` and check no errors (0 errors in the modified file; the full-package run surfaces 37 pre-existing errors across 16 unrelated files, none touched by this change)
- [ ] I have made corresponding changes to the documentation (not applicable — no public API or docs changed, test-only fix)


deepchem/metrics/tests/test_genomics.py
```diff
@@ -12,6 +12,12 @@
 except:
     has_tensorflow = False
 
+try:
+    import simdna  # noqa: F401
+    has_simdna = True
+except ModuleNotFoundError:
```
siddhant-shahhh: Small thing: ModuleNotFoundError is a subclass of ImportError, so this catches the
"simdna isn't installed" case but not the "simdna is installed but its own imports
fail" one — that raises plain ImportError and would escape the guard, leaving the
test erroring instead of skipping.

The equivalent guards in #5051 use `except ImportError`. Worth matching?

siddhant-shahhh (commented):
Thanks for picking this up so quickly.

I filed the issue, so I had the environment sitting there — checked your branch out
and ran it on macOS 26.6.2, Python 3.12.13, with neither TensorFlow nor simdna
installed.

On master:

    3 failed, 1 passed

On this branch:

    test_get_motif_scores                SKIPPED
    test_get_pssm_scores                 PASSED
    test_in_silico_mutagenesis_nonzero   SKIPPED
    test_in_silico_mutagenesis_shape     SKIPPED
    1 passed, 3 skipped

deepchem/metrics and deepchem/splits together: 58 passed, 3 skipped, nothing newly
broken.

Confirms the fix from my end. One small question left inline.

deepchem/metrics/tests/test_genomics.py
sumitjhadev: Good catch on the except ImportError fixed. While I was in there, I also noticed the guard only checked import simdna, not from simdna import synthetic, which is what get_motif_scores() actually needs internally, so I added that too. Re-tested both with and without the deps installed, all good on my end. Thanks for digging into this with your own environment that was a solid check.

Would appreciate another look when you get a chance.

siddhant-shahhh (approved):
Thanks for the change re-ran on the same setup (no TensorFlow, no simdna) and get
1 passed, 3 skipped, with metrics and splits clean. Guarding the ​from simdna import synthetic​ line as well as the module import is better than what I suggested, since
get_motif_scores needs both. LGTM.

Shriprasad-P (commented):
Thanks for opening this. The approach looks reasonable. I'd mainly want confirmation that existing callers keep working and that tests cover the failure path.

PR #4763, opened 2026-02-22
Fix: Implement shard_size in DataLoader shard_generator methods

saurabhhhcodes:
Fixes #4741. Added shard_size support in FASTALoader, SAMLoader, BAMLoader, and CRAMLoader.

PR #5133, opened 2026-09-13
Fix FASTALoader discarding sequences when a record is empty

### LINKED ISSUE #5132: FASTALoader silently discards sequences when a FASTA record is empty
## 🐛 Bug

FASTALoader (legacy=False) silently returns fewer sequences than the input file
contains when any record has no sequence body. No error and no warning is raised 
the sequences are just missing from the resulting dataset.

Two cases:

1. An empty record in the middle of the file discards every sequence read before it.
2. A file ending with a header and no sequence returns an empty dataset.

Both are plausible in practice: FASTA files produced by filtering steps often contain
empty records, and a truncated download can end on a header line.

## To Reproduce

1. Install DeepChem from source on master (455d07f3e) with `pip install -e .`
2. Run the following:

```python
import tempfile, os, deepchem as dc

cases = [
    ("empty record mid-file", ">seq0\nAAAA\n>seq1\n>seq2\nCCCC\n", 2),
    ("file ends on a header", ">seq0\nAAAA\n>seq1\n", 1),
]

for name, content, expected in cases:
    with tempfile.TemporaryDirectory() as d:
        path = os.path.join(d, "t.fasta")
        open(path, "w").write(content)
        got = dc.data.FASTALoader(legacy=False).create_dataset(path).X.shape[0]
        print(f"{name}: expected {expected}, got {got}")
```

Output:

    empty record mid-file: expected 2, got 1
    file ends on a header: expected 1, got 0

## Expected behavior

Empty records should be skipped, leaving the sequences around them intact. An empty
record should not cause previously read sequences to be dropped, and it should not
cause the loader to return an empty dataset.

## Environment

* OS: macOS 26.6.2 (arm64)
* Python version: 3.12.13
* DeepChem version: 2.8.1.dev (editable install from master @ 455d07f3e)
* RDKit version (optional): 2026.3.6
* TensorFlow version (optional): not installed
* PyTorch version (optional): 2.14.0
* Any other relevant information: the non-legacy path is affected; legacy=True uses
  BioPython via encode_bio_sequence and is not involved.

## Additional context

I have a fix and tests ready and will open a PR shortly.

siddhant-shahhh:
## Description

Fix #5132

`_add_sequence` in `FASTALoader._read_file` returns an empty array when given an
empty sequence:

    if sequence is None or len(sequence) <= 0:
        return np.array([])

Because the return value is assigned back to `sequences`, this doesn't just skip the
empty record it discards everything accumulated so far. Harmless on the first call,
when `sequences` is already empty, but not afterwards.

Two consequences, both silent:

1. An empty record mid-file discards every sequence read before it.

       >seq0
       AAAA
       >seq1      <- empty
       >seq2
       CCCC

   Returns 1 sequence on master; should be 2.

2. A file ending with a header and no sequence returns an empty dataset, because the
   final `_add_sequence` call wipes everything.

       >seq0
       AAAA
       >seq1

   Returns 0 sequences on master; should be 1.

The fix returns `sequences` unchanged. Two regression tests are added; both fail on
master and pass with the change. Existing behaviour is unaffected — example.fasta and
multi-line sequences give identical results before and after, and the full
deepchem/data/tests failure set is identical on master and this branch (29 either
way, all pre-existing missing-optional-dependency failures).

There is an identical `_add_sequence` in the FASTQ loader, but it is always passed a
length-1 array, so that branch never fires. I've left it alone rather than widen the
diff.

Note: #2589, #4743, #4763 and #4904 all touch this region for sharding or style
reasons, none of them changing this line. Happy to rebase around whichever lands.

No dependencies added.

## Type of change

## Checklist

  - [x] Run `yapf -i <modified file>` and check no errors (yapf version 0.32.0)
- [ ] I have made corresponding changes to the documentation — n/a, no documented behaviour changes


Shriprasad-P (commented):
Appreciate the cleanup here. Change set reads well — please make sure docs/changelog (if required by the repo) stay in sync with the behavior change.

PR #5134, opened 2026-09-14
Correct Typo initalizer -> initializer in DTNNEmbedding

pankajyadavbuildsai99-tech:
This PR fixes the misspelled parameter initalizer to initializer in DTNNEmbedding class (deepchem/models/torch_models/layers.py).

- Renamed parameter to correct spelling
- Added backward compatibility: old initalizer kwarg still works with DeprecationWarning
- Fixed docstring and _repr_

Fixes typo that could confuse users.


Shriprasad-P (commented):
Solid direction. One thing to verify: thread-safety / concurrent use if this touches shared state, plus a regression test for the reported case.

pankajyadavbuildsai99-tech:
> Solid direction. One thing to verify: thread-safety / concurrent use if this touches shared state, plus a regression test for the reported case.

Hi @reviewer, thanks for the suggestion!

I've added backward compatibility for the old misspelled kwarg:

- Fixed typo `initalizer` -> `initializer` in `DTNNEmbedding`
- Added support for both spellings with `DeprecationWarning` for old spelling
- Added regression test `test_dtnn_embedding_initializer_backward_compat` in `test_dtnn_embedding_typo.py` to verify both spellings work and warning is raised

The old code using `initalizer` will now still work but show a deprecation warning, and new code can use correct `initializer`. Let me know if any other changes needed!

PR #5136, opened 2026-09-16
Preserve data types when resharding DiskDataset

Shubham-Padkonde:
## Description

`DiskDataset.reshard()` initializes its feature, label and weight buffers as
float64. Concatenating the first shard with these empty buffers promotes
float32, integer and Boolean arrays to float64. For example, the int64 value
9007199254740993 becomes 9007199254740992.0 after resharding. This also doubles
the storage required by float32 data.

Initialize the empty buffers from slices of the first shard so they retain
the input dtypes. Subsequent shards continue to use NumPy's normal type
promotion rules. Existing handling of unlabeled datasets is preserved.

Add regression coverage for float32, int64, uint64 and Boolean arrays, including
integers beyond float64's exact range. The tests split and join uneven input
shards with output shard sizes 3, 1 and 7 and verify features, labels, weights,
IDs and shard counts.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation-only change

## Validation

- All four new regression cases fail on upstream commit
  `455d07f3e17e880a4d980b5b0365e4a02b417e14` and pass with this change.
- 54 relevant dataset tests pass. Four framework-adapter tests were excluded
  because TensorFlow and PyTorch are not installed in this CPU core environment.
- YAPF 0.32.0 formatting, doctests for both modified files, and `git diff --check`
  pass. The modified test file passes flake8.
- Whole-file flake8 reports six existing E226 findings outside the changed code.
  `mypy -p deepchem` reports 37 existing errors in 16 other files. Both were
  checked against the unmodified source with the same dependencies.
- Python 3.12.14 and NumPy 1.26.4. The full project suite and optional framework
  integrations were not run.

Prepared with Codex assistance for investigation, implementation and validation.
Searches for `reshard` and `dtype DiskDataset` found no matching current fix.


siddhant-shahhh (approved):
Had a look at this one since reshard gets used a lot and a silent dtype change is easy to miss.

The bug is real, and it's worse than cosmetic. The generator seeds its buffers with
np.zeros((0,) + shape), which is float64, so the first np.concatenate promotes every shard.
Simulating the generator on a 7-row, 2-shard dataset:

    dtype     before     after
    float32   float64    float32
    int64     float64    int64
    uint64    float64    uint64
    bool      float64    bool

Shard counts come out identical either way, so nothing else shifts.

And above 2**53 the promotion doesn't only change the dtype, it changes the values, which is
exactly what the += 2**53 offset in the test catches. Nice test.

One thing the fix doesn't cover, not a blocker: on the `y is None` path, y and w are assigned
from the existing float64 buffers before the new guard runs, so label-less datasets still come
out float64 for y/w. Might be worth a line of comment so the next reader doesn't take it for an
oversight.

yapf 0.32.0 and flake8 are both clean on the two files.

LGTM.

